PR #1, opened 2016-10-26
Tep16

cpascual:
This Pull request aims to introduce the [TEP16](https://github.com/cpascual/taurus/blob/tep16/doc/source/tep/TEP16.md):

```
Title: Moving Taurus to Github
Abstract: Move Taurus project from its current hosting in SourceForge to 
 GitHub. The move affects the code repository, the ticket tracker and the 
 wiki pages. It also proposes to change the contribution and the TEP 
 workflow to make use of the Pull Request feature.
```
## Roadmap for this TEP:
- Since the "rough consensus" on the convenience of migrating has already been reached during the last Sardana Workshop, I plan to move the state of TEP16 to CANDIDATE  before the end of October 2016
- Then, I would like to further advance in the implementation of the remaining changes so that we can test them and vote if we approve it during November 2016


LJBD:
As I said during the Sardana Workshop: for me an official migration to GitHub is a must.

There's one thing that needs a little more clarification IMHO - which state a TEP is in currently. I think that labels could be used to tag a TEP PR in order to mark the current state of the corresponding TEP. Therefore, a set of labels for TEP states would have to be created.

What do you think about it?

cpascual:
> There's one thing that needs a little more clarification IMHO - which state a TEP is in currently.

In principle the state is one of the things stated in the TEP header.

> I think that labels could be used to tag a TEP PR in order to mark
> the current state of the corresponding TEP. 

One important thing about the TEPs is that they are self-contained text documents (e.g. they state the URL, the date, the author, etc in their header). 
So, I propose to use labels as you suggest, but not as a "mandate" of TEP16. That is, the authoritative source of the STATE of a TEP should still be the header of the TEP document. and the labels will be used just as some optional "sugar" on which we do not depend.
Similarly, the TEP index page may make use of the labels to help in the classification.

Do you agree?

LJBD:
This sounds reasonable. I guess that adding labels would be a good practice, it doesn't have to be considered mandatory.

cpascual:
(This is a trasncription of a mail sent to tauruslib-users@lists.sourceforge.net )

```
Subject: [Tauruslib-users] TEP16 - Voting for CANDIDATE --> Accepted
From:	Me (CELLS)
To:	tauruslib-users@lists.sourceforge.net
Date:	11/16/16 6:41 PM
```
Hi all,

Finally the CANDIDATE implementation of the TEP16 (migration of Taurus to 
GitHub) is ready to be reviewed.

Please review the proposed changes in https://github.com/taurus-org/taurus/pull/1

... and discuss in that same Pull Request if you have any comment

Also, a representative from each institute should cast a vote regarding 
whether we should accept the TEP16 (please do it ASAP by commenting on the 
Pull Request itself).

Note that ACCEPTING TEP16, automatically moves TEP0 and TEP7 to OBSOLETE 
state.

Cheers,

Carlos

PD: please note that the "service-agnostic URL for the TEPs" mentioned in the 
TEP16 will only work once we actually merge the Pull Request (i.e., do not 
worry about broken links until then)

PPD: in the eventuality that we do not accept the TEP16 , I would undo all the 
changes I did on SourceForge... but I hope that it is not necessary! ;)

teresanunez:
Hi Carlos, thanks for the work. We accept the changes you propose in the TEP16, so you have
 a YES from DESY for going ahead.

LJBD:
Of course it's also a YES from SOLARIS!

tiagocoutinho:
Nice work! Fine with me too. What is the official way to vote? we just put a thumbs up on your initial request or we write a message like this

amilan:
Hi all, 

It's also a big YES from MAX IV.

Cheers.

cpascual:
> What is the official way to vote? we just put a thumbs up on your initial request or we write a message like this?

We didn't define it. I asked for a comment, so let's do it comment. Thumbs up may be more GitHubish...

PR #8, opened 2016-11-02
Deprecate taurus.external.ordereddict

cpascual:
As a first step to fix bug-271, add a deprecation warning
to the init of taurus.external.ordereddict


cpascual:
@taurus-org/integrators Please, someone integrate this

PR #9, opened 2016-11-02
Fix Bug 343 (support spyder>=3)

cpascual:
Fix bug from https://sourceforge.net/p/tauruslib/tickets/343/

Attempt to import spyder and fall back to spyderlib if there is an exception.

### DIFF
--- a/lib/taurus/core/taurushelper.py
+++ b/lib/taurus/core/taurushelper.py
@@ -186,10 +186,15 @@ def __get_qtcontrols_version_number():
 
 def __get_spyderlib_version():
     try:
-        import spyderlib
-        return spyderlib.__version__
+        # spyderlib got renamed to spyder in v3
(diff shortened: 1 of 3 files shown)


reszelaz:
I have reviewed this PR and manually tested it with spyder v3.
Since we have dropped support to spyderlib we need to upgrade the taurus-test docker container before accepting this PR.

cpascual:
I just merged a PR in the docker image repo for this: https://github.com/cpascual/taurus-test/pull/1

cpascual:
After some trouble with the docker image build, now the image should be ok, so I retriggered the travis build

cpascual:
I ran `python <taurus>/qt/qtgui/editor/tauruseditor.py` in a taurus-test docker with this installed file and it works ok for me. So IMHO, this could be merged

PR #347, opened 2016-11-16
Add bumpversion support (and bump version to 4.0.2-dev)

cpascual:
Add [bumpversion](https://github.com/peritus/bumpversion) support to taurus. This allows us to keep a consistent version scheme.

With this, the release process would be as follows (it can be done manually or eventually be automated in the Continuous Delivery process):

- We start in develop, with an unreleased version (i.e., with "-dev" suffix). For example, say "4.0.2-dev"
- start release branch with: 
    ```
    git checkout -b release-`bumpversion --dry-run --list release | grep new_version= | sed -r s,"^.*=",,` develop
    ```
   (the branch will be called  "release-4.0.2")

- call `bumpversion release` (version in the release branch will now be "4.0.2")
- work on release branch and commit (if needed)
- edit CHANGELOG.md (if needed)
- merge release branch into master and tag with:
  ```
  git tag `bumpversion --dry-run --list release | grep current_version= | sed -r s,"^.*=",,`
  ```
- merge release branch into develop and bump patch version with `bumpversion patch`(version in develop will now be "4.0.3-dev")
- During the coming development process, any change that adds a new feature should call `bumpversion minor` and any change that produces backwards incompatibility should call `bumpversion major`. Bug fixes do not need to bump the version since we already did just after the release.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,21 @@
+[bumpversion]
+commit = True
+message = Bump version {current_version} to {new_version}
+tag = False
+tag_name = {new_version}
+current_version = 4.0.2-dev
+parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+))?
+serialize = 
+	{major}.{minor}.{patch}-{release}
+	{major}.{minor}.{patch}
+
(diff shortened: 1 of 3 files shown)


cpascual:
@taurus-org/integrators Is anybody looking into integrating this one?

PR #348, opened 2016-11-22
Remove from Taurus objects the direct Logger dependence (TEP8)

cmft:
Remove from Taurus objects the direct Logger dependence

### DIFF
--- a/lib/taurus/core/evaluation/evalattribute.py
+++ b/lib/taurus/core/evaluation/evalattribute.py
@@ -289,7 +289,7 @@ class EvaluationAttribute(TaurusAttribute):
         try:
             v = evt_value.rvalue
         except AttributeError:
-            self.trace('Ignoring event from %s' % repr(evt_src))
+            self.debug('Ignoring event from %s' % repr(evt_src))
             return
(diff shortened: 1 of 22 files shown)


cpascual:
This is currently breaking the tests. Please have a look at:
https://travis-ci.org/taurus-org/taurus/builds/177951399

cpascual:
Note that the current implementation is not backwards compatible in general (not all of the members of the Logger class previously available are accessible in the new implementation).

So, either we find a way to code a fully compatible implementation, or we should re-evaluate the gains from this TEP.

cmft:
I agree. I did an investigation about it, and I think that I found an option to do not break the old code.
See it [here](https://gist.github.com/cmft/5994c59d9614fdb4dee3d50754e203a4). What do you think @cpascual?

cpascual:
After talking to @cmft , we decided to "stop pretending" and close this PR. The TEP8 would need a full reimplementation, but all work on that area has been stalled for too long since it fell from our priority list. We may re-take the implementation of TEP8 at some later time if it goes up in our (or someone else's) priority list again.

PR #351, opened 2016-12-01
Make Project page point to the repo, not issues

cpascual:
The Project Page link in the doc index (only visible on the left bar in RTD)
points to the issues page. It should poit to the root repo page instead.
Change it


reszelaz:
I agree with the change. I merged it regardless of the appveyor errors - not related to this PR.

PR #356, opened 2016-12-13
Fix MSI creation

### LINKED ISSUE #352: appveyor build fails due to problematic version string
The problem seems to be when building the msi.

See https://ci.appveyor.com/project/taurusorg/taurus/build/1.0.4

```
python setup.py bdist_wheel bdist_msi
(...)
C:\projects\taurus\venv_build\lib\site-packages\setuptools\dist.py:331: UserWarning: Normalizing '4.0.2-dev' to '4.0.2.dev0'
  normalized_version,
  File "setup.py", line 161, in <module>
    test_suite='taurus.test.testsuite.get_taurus_suite',
  File "c:\python27-x64\Lib\distutils\core.py", line 151, in setup
    dist.run_commands()
  File "c:\python27-x64\Lib\distutils\dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "c:\python27-x64\Lib\distutils\dist.py", line 972, in run_command
    cmd_obj.run()
  File "c:\python27-x64\Lib\distutils\command\bdist_msi.py", line 236, in run
    sversion = "%d.%d.%d" % StrictVersion(version).version
  File "c:\python27-x64\Lib\distutils\version.py", line 40, in __init__
    self.parse(vstring)
  File "c:\python27-x64\Lib\distutils\version.py", line 107, in parse
    raise ValueError, "invalid version number '%s'" % vstring
ValueError: invalid version number '4.0.2.dev0'
```

cpascual:
The introduction of proper semver support with bumpversion (#347)
broke the creation of MSI packages (see #352). The reason is that
distutils.version.StrictVersion does not support semantic versioning
for pre-releases. As a workaround, normalize the version to be
acceptable for StrictVersion. This Fixes #352.

Note that the normalized version is identical to the semver except for
pre-releases (where '-dev' is replaced by 'a0').

UPDATE: a different approach was finally selected after this suggestion (https://github.com/taurus-org/taurus/pull/356#discussion_r92418123) by @reszelaz

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -137,9 +137,15 @@ classifiers = [
     'Topic :: Software Development :: Widget Sets',
 ]
 
+# Unfortunately, distutils.version.StrictVersion does not support semantic
+# versioning (www.semver.org) for pre-releases. This breaks, e.g,  bdist_msi
+# As a workaround, we normalize the version to be acceptable for StrictVersion
+# Note that the normalized version is identical to the semver except for
+# pre-releases (where '-dev' is replaced by 'a0').
+_version = release.version.replace('-dev', 'a0')
 
 setup(name='taurus',
-      version=release.version,
+      version=_version,
       description=release.description,
       long_description=release.long_description,
       author=release.authors['Tiago_et_al'][0],
--- a/setup.py
+++ b/setup.py
@@ -137,15 +137,9 @@ classifiers = [
     'Topic :: Software Development :: Widget Sets',
 ]
 
-# Unfortunately, distutils.version.StrictVersion does not support semantic
-# versioning (www.semver.org) for pre-releases. This breaks, e.g,  bdist_msi
-# As a workaround, we normalize the version to be acceptable for StrictVersion
-# Note that the normalized version is identical to the semver except for
-# pre-releases (where '-dev' is replaced by 'a0').
-_version = release.version.replace('-dev', 'a0')
 
 setup(name='taurus',
-      version=_version,
+      version=release.version,
       description=release.description,
       long_description=release.long_description,
       author=release.authors['Tiago_et_al'][0],
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,7 +3,7 @@ commit = True
 message = Bump version {current_version} to {new_version}
 tag = False
 tag_name = {new_version}
-current_version = 4.0.2-dev
+current_version = 4.0.3-alpha
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\-(?P<release>[a-z]+))?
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -137,9 +137,15 @@ def get_release_info():
     'Topic :: Software Development :: Widget Sets',
 ]
 
+# Unfortunately, distutils.version.StrictVersion does not support semantic
+# versioning (www.semver.org) for pre-releases. This breaks, e.g,  bdist_msi
+# As a workaround, we normalize the version to be acceptable for StrictVersion
+# Note that the normalized version is identical to the semver except for
+# pre-releases (where '-dev' is replaced by 'a0').
+_version = release.version.replace('-dev', 'a0')
```
reszelaz: Isn't it possible to change it directly on the bumpversion configuration level?

cpascual: > Isn't it possible to change it directly on the bumpversion configuration level?

It is definitely possible (and easy). But I will try to justify my decision: before implementing my proposed solution I considered the following options:

1.- (what @reszelaz suggests) use the distutils version style (e.g. "4.0.3a0") instead of the semver style ("4.0.3-dev") in the `taurus.core.release.version` and configure bumpversion to update it directly. This has the advantage of not differentiating between the taurus **module** version and the taurus **distribution** version. But I rejected it because that would mean giving up on semver, which I think it is not worth.

2.- limit the scope of the normalization of the version to the bdist_msi command. This can be done by monkey patching distutils from setup.py (see [this](http://bugs.python.org/issue6040#msg133094)). I rejected this approach because it implies polluting our setup.py with potentially fragile code and it results in an inconsistent version naming between different  distributions of the same source (e.g., the .egg, .wheel or .tar.gz would use semver while the .msi would use distutils style)
 
The main drawback of my proposed solution is that it makes a difference between the **module** version and the **distribution** version (i.e., after installing **a prerelease** taurus, you would get a different answer if you interrogate `taurus.Release.version` than if you interrogate `pkg_resources.require("taurus")[0].version`). But note that this is the case already, and that this only happens on pre-release installations, and can be understood if you consider that module != distribution

reszelaz: After performing some tests we have observed that `X.Y.Z-alpha` is normalized by distutils to a MSI compatible version number `X.Y.Za0`. Using this is still compatible with *semver* and does not require any special replacements for the `distutils`. Another option would be to use `X.Y.Z-a0` that would also normalize to `X.Y.Za0`.

cpascual: Nice!: I am going to change the implementation to use -alpha instead of -dev

cpascual:
The new implementation changes -dev to -alpha and allows to avoid the manual normalization of the version in setup.py (distutils already supports the new semver-compatible version)

PR #362, opened 2017-01-11
Fix #359 support of read-only uchar tango attrs

### LINKED ISSUE #359: Taurus4 regression: cannot read TangoAttribute uchar_image_ro
Hi,

On Windows machine, with Taurus 4.0.1 

When settting the model of a TaurusImageDialog with an image of type uint8, it failed to show the image.

The problem seems to be in the taurusvaluestable.py, getting 'rvalue is invalid' Exception in setModel method. 

It can be reproduces using the TangoTest device.

cpascual:
Fix regression in Taurus4 that broke TangoAttribute.read() for read-only uchar Tango attributes.
This PR fixes #359.


cpascual:
@taurus-org/integrators , can you please integrate this?

(see conversation in #359)

teresanunez:
Carlos, I have tested it (applying the patch to the current develop branch of taurus) and if I display the
 uchar_ro array with:

 taurusform 'sys/tg_test/1/uchar_image_ro'

 I do not see the right characters. It shows charachters (before applying the patch does not) but
 they are not digits (like they are using taurus 3.x)

cpascual:
Hi @teresanunez, thanks for reviewing... See #367 (I will push some changes to this PR to revert to what was shown in Taurus3)

teresanunez:
Hi Carlos, should I review this now, or the problem with Travis has to be solved first?.

cpascual:
> Hi Carlos, should I review this now, or the problem with Travis has to be solved first?.

I think that the problem with travis **in this PR** is a spurious one (related to #261). I just triggered a rebuild manually. Therefore, you can review it. Thanks!

cpascual:
Confirmed. The travis error was spurious. You may go ahead with the review.

teresanunez:
Hi Carlos, now I see an image, not an array of integers like in taurus3. For me is ok, but
 is that what you want?.

cpascual:
Yes, that is just a matter of what is the **default** widget that the TaurusForm uses. IMHO, it makes more sense to show numeric arrays as images by default. Note that you can force TaurusForm  to show the values in a table if you right-click on the label and choose `change read widget-->TaurusValuesTable`

Thanks again for reviewing

PR #380, opened 2017-01-24
Initialize X with thread-safety when using TaurusApplication.

cpascual:
Sometimes (e.g. running `taurusgui example01`) one gets the following:
 [xcb] Unknown request in queue while dequeuing
 [xcb] Most likely this is a multi-threaded client and XInitThreads has
       not been called
 [xcb] Aborting, sorry about that.

 Fix according to http://stackoverflow.com/a/31967769 , by setting:
 `Qt.QCoreApplication.setAttribute(Qt.Qt.AA_X11InitThreads)`


cpascual:
@taurus-org/integrators can someone please integrate this?

teresanunez:
I will do it.

PR #384, opened 2017-01-25
 Fix #382 TangoFactory ignores the default_tango_host variable

### LINKED ISSUE #382: TangoFactory  ignores the default_tango_host variable
If you have defined the env. var TANGO_HOST in any of the possible ways in your system but you decide to set  a default_tango_host for your application (via TaurusManager or TangoFactory). 
TangoFactory ignores the set default_value, when you ask for a Device without define the authority part. It uses the value defined by the TANGO_HOST env. var. instead of the set _default_tango_host variable.

e.g. 

If your TANGO_HOST = FOO:10000
``` python
import taurus
f = taurus.Factory()
f.set_default_tango_host('BAR:10000')
d = f.getDevice('a/b/c')
# It prints tango://FOO:10000/a/b/c instead of tango://BAR:10000/a/b/c
print d.getFullName() 
```

cmft:
TangoFactory ignores the set default_value, when you ask for a
Device without define the authority part. It uses the value defined
by the TANGO_HOST env. var. instead of the set _default_tango_host
variable.

The main problem comes from the TangoDevice validator that does not
ask for the set TangoFactory._default_tango_host variable.

Fixes #382

### DIFF
--- a/lib/taurus/core/tango/tangofactory.py
+++ b/lib/taurus/core/tango/tangofactory.py
@@ -162,10 +162,11 @@ class TangoFactory(Singleton, TaurusFactory, Logger):
         return dict(self.tango_db)
 
     def set_default_tango_host(self, tango_host):
-        """Sets the new default tango host.
+        """Sets the new default tango host. The method will transform the given
+        name to an Authority URI.
 
-        :param tango_host: (str) the new tango host or None to use the defined
-        by $TANGO_HOST env. var
+        :param tango_host: (str) the new tango host. It accepts any valid Tango
+        authority name or None to use the defined by $TANGO_HOST env. var.
         """
         # Translate to Authority URI
         if tango_host and "//" not in tango_host:
--- a/lib/taurus/core/tango/tangofactory.py
+++ b/lib/taurus/core/tango/tangofactory.py
@@ -176,6 +176,7 @@ class TangoFactory(Singleton, TaurusFactory, Logger):
         """Retruns the current default tango host
         """
         return self._default_tango_host
+
     def registerAttributeClass(self, attr_name, attr_klass):
         """Registers a new attribute class for the attribute name.
 
@@ -286,16 +287,20 @@ class TangoFactory(Singleton, TaurusFactory, Logger):
            :raise: (taurus.core.taurusexception.TaurusException) if the given
                    dev_name is invalid.
         """
+        validator = _Device.getNameValidator()
+        groups = validator.getUriGroups(dev_name)
+
+        if groups is None:
+            raise TaurusException("Invalid Tango device name '%s'" % dev_name)
+
         d = self.tango_devs.get(dev_name)
         if d is None:
             d = self.tango_alias_devs.get(dev_name)
-        if d is not None:
-            return d
-
-        validator = _Device.getNameValidator()
-        groups = validator.getUriGroups(dev_name)
-        if groups is None:
-            raise TaurusException("Invalid Tango device name '%s'" % dev_name)
+            if d is not None:
+                auth = validator.getUriGroups(d.getFullName()).get('authority')
+                # Check is the alias belongs to the same authority
+                if auth == groups.get('authority'):
+                    return d
 
         full_dev_name, _, _ = validator.getNames(dev_name)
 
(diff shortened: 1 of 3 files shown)


cpascual:
Hi @cmft, do you think that this may also solve the problem reported by @jkotan in https://github.com/taurus-org/taurus/issues/79#issuecomment-274733195 ?

cmft:
Hi @cpascual, from the point of view of taurus I think so but not 100% sure. Nevertheless sardana still need  some adaptations related with the uses of "normal names" instead of fullnames.

lib/taurus/core/tango/tangovalidator.py
```diff
@@ -90,19 +90,22 @@ def getNames(self, fullname, factory=None, queryAuth=True):
         if groups is None:
             return None
 
-        import PyTango
-        default_authority = '//' + PyTango.ApiUtil.get_env_var('TANGO_HOST')
+        default_authority = None
```
cpascual: This is breaking the API that says that the authority is not queried unless `queryAuth` is True

lib/taurus/core/tango/tangofactory.py
```diff
@@ -166,9 +166,16 @@ def set_default_tango_host(self, tango_host):
...
+        # Translate to Authority URI
+        if "//" not in tango_host:
+            tango_host = "//{0}".format(tango_host)
         self._default_tango_host = tango_host
         self.dft_db = None
 
+    def get_default_tango_host(self):
+        """Retruns the current default tango host
+        """
+        return self._default_tango_host
```
cpascual: (minor) missing blank line

lib/taurus/core/tango/tangofactory.py
```diff
@@ -286,17 +287,21 @@ def getDevice(self, dev_name, create_if_needed=True, **kw):
...
         validator = _Device.getNameValidator()
         groups = validator.getUriGroups(dev_name)
+
         if groups is None:
             raise TaurusException("Invalid Tango device name '%s'" % dev_name)
 
+        d = self.tango_devs.get(dev_name)
+        if d is None:
+            d = self.tango_alias_devs.get(dev_name)
+            if d is not None:
```
cpascual: Shouldn't this block be unindented?
(I just say it by comparison with the original code)

cpascual:
After a long offline discussion with @cmft we agreed in changing the approach regarding the handling of the cached aliases by the TangoFactory:

our proposal is to leave the `TangoFactory.getDevice` code as it was originally and just force a wipe of `TangoFactory.tango_alias_devs`  when `TangoFactory.set_default_tango_host()` is called (i.e., if the default TANGO HOST changes, the cache of known aliases is no longer valid).

lib/taurus/core/tango/tangovalidator.py
```diff
@@ -90,19 +90,22 @@ def getNames(self, fullname, factory=None, queryAuth=True):
...
+        default_authority = factory.get_default_tango_host()
+
+        if default_authority is None:
+            import PyTango
+            default_authority = "//" + PyTango.ApiUtil.get_env_var('TANGO_HOST')
 
         authority = groups.get('authority')
         if authority is None:
-            groups['authority'] = authority = default_authority
+            groups['authority'] = authority = format(default_authority)
```
cpascual: Whut using `format` here?

lib/taurus/core/tango/tangofactory.py
```diff
@@ -164,11 +164,22 @@ def getExistingDatabases(self):
     def set_default_tango_host(self, tango_host):
         """Sets the new default tango host.
```
cpascual: The docstring should also mention that the device alias cache is cleared

lib/taurus/core/tango/tangofactory.py
```diff
@@ -164,11 +164,22 @@ def getExistingDatabases(self):
...
 
-        :param tango_host: (str) the new tango host
+        :param tango_host: (str) the new tango host or None to use the defined
```
cpascual: The docstring should also mention that the device alias cache is cleared

lib/taurus/core/tango/tangovalidator.py
cmft: It was a copy/paste error. I'll fix it.

PR #388, opened 2017-01-26
Fix #387 UnboundLocalError in TangoAttrValue

### LINKED ISSUE #387: Handling Tango attributes of invalid quality
Hello!

I've stumbled across a possible bug (or unhandled exception): when creating a TaurusLabel and setting its model to a certain attribute that has invalid quality, I've received the following exception:
```python
push_event generated the following python exception:
  File "/usr/lib/python2.7/site-packages/taurus/core/tango/tangoattribute.py", line 712, in push_event
    event.attr_value), None
  File "/usr/lib/python2.7/site-packages/taurus/core/tango/tangoattribute.py", line 409, in decode
    value = TangoAttrValue(pytango_dev_attr=attr_value, attr=self)
  File "/usr/lib/python2.7/site-packages/taurus/core/tango/tangoattribute.py", line 103, in __init__
    p.value = numpy.empty(shape, dtype=dtype)
UnboundLocalError: local variable 'shape' referenced before assignment
```

In order to investigate, I've created a simple Device with only one attribute and a command which changes the attribute's validity. Unfortunately, I haven't been able to see that exception again.
What I have discovered, though, is a possible regression: when an attribute's quality changes while there's a TaurusLabel displaying it, Taurus 3 starts to show a grey area. Taurus 4 continues to show the last value (see attached screenshot).
taurus_validity_test
I guess that Taurus 3 behaviour is better since it lets a user know that something is wrong with an attribute.

So, to sum things up, there are 2 issues mentioned:
1. A weird error appearing for only one attribute with invalid quality.
2. A possible regression in displaying invalid attributes' values.

Best regards,
Łukasz Dudek

cmft:
PyTango.DeviceAttribute.isempty returns True with scalar values
and Taurus trusts in this value without check the data_type.

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -96,7 +96,8 @@ class TangoAttrValue(TaurusAttrValue):
         if p.has_failed:
             self.error = PyTango.DevFailed(*p.get_err_stack())
         else:
-            if p.is_empty:  # spectra and images can be empty without failing
+            # spectra and images can be empty without failing
+            if p.is_empty and self._attrRef.data_format != DataFormat._0D:
                 dtype = FROM_TANGO_TO_NUMPY_TYPE.get(
                     self._attrRef._tango_data_type)
                 if self._attrRef.data_format == DataFormat._1D:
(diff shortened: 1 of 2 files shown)


cpascual:
According to https://github.com/taurus-org/taurus/issues/387#issuecomment-276296348 this fixes the issues. Merging

Issue #392, opened 2017-01-31
Wrong formatting of Tango attributes


LJBD:
Hello!

Taurus 3.x used to read format from attribute configuration in Tango and display it accordingly. Taurus 4.0.x does not do that.
I wrote 1.123456789 to float_scalar attribute of a TangoSchemeTest device. The configured format is %6.2f. You can see behaviour of Taurus 3.7.3, 4.0.1, 4.0.3 and 4.0.4-alpha (I used Carlos Falcon's bug-387 branch).
![taurus_display_format_test](https://cloud.githubusercontent.com/assets/9516795/22457124/664eba5c-e796-11e6-8df5-abb8064db597.JPG)

cmft:
I can reproduce it. Thanks!

cpascual:
If I remember correctly, we left this out of Taurus4 knowingly because doing a proper (scheme-agnostic) support for formatting in taurus required some deep discussion. I think that we could now open such discussion (maybe it is even worth a TEP).

See https://github.com/taurus-org/taurus/pull/393#issuecomment-276603663:

>We should have a scheme-agnostic property for specifying the format (I propose the "new" standard python string-formating better than old style c-like formatting because it would integrate more naturally with quantities), and encode/decode it from whatever an scheme uses.

cmft:
Hi @cpascual , @LJBD ,
I think we can fix it, with minor API changes (as temporally solution) and assuming that taurus  uses the "new" standard python string-formatting. The only scheme affected for the moment is Tango.  I implemented a solution in one of my branches https://github.com/cmft/taurus/tree/bug-392.

What do you think? Is it enough?

LJBD:
It looks all right from a user point of view. I don't feel versed in taurus enough to be able to asses if the code is written well, though.
Anyway, I've noticed a funny behaviour: I've launched a ATK Attribute property editor and a TaurusLabel for the same attribute. When I'm changing format and clicking "Apply changes", the TaurusLabel updates instantly. When I'm changing units, it does not - it updates units only when value of the attribute changes. Is that an expected behaviour?

cpascual:
> What do you think? Is it enough?

I did some comments on the commits of [your branch]( https://github.com/cmft/taurus/tree/bug-392), but I think that we should stop and discuss this more before actually looking into implementations

Here are some points that I think should be clarified:

- Does a "format" really belong to the Model? IMHO it is clearly something related to the view and not the Model (e.g. 2 different widgets may require different formats). 
[...]

PR #393, opened 2017-01-31
(WIP) Fix bug#392 Wrong formatting of Tango attributes

cmft:
Taurus 4 widgets ignore the format of the represented attributes (see #392).
This is an introduce regression after the Taurus4 refactoring.

Fix it in TaurusBaseComponent.displayValue method, now the represented
value respect the attribute format.


cmft:
It fails with quantities. I'm working on it. Do not merge!!

cpascual:
@cmft: I think that we should consider a different approach.
We should have a scheme-agnostic property for specifying the format (I propose the  ["new" standard python string-formating](https://docs.python.org/2/library/string.html#format-string-syntax) better than old style c-like formatting because it would integrate more naturally with quantities), and encode/decode it from whatever an scheme uses.

cmft:
I agree @cpascual, this also simplify  the implementation to fix the bug. I will close this PR and I will open a new one with other implementation.

PR #395, opened 2017-02-01
Fix bug#394 AssertionError in TangoAttribute methods

cmft:
The TangoAttribute setRange, setAlarms, and setWarnings methods
raise an AssertionError when the attribute has non dimensionless unit
and any of the input parameter does not define the unit. This a
breakup of the back. comp.

Fix it, setting the attribute unit to the dimensionless input.

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -811,7 +811,11 @@ class TangoAttribute(TaurusAttribute):
             limits = limits[0]
         low, high = limits
         low = Quantity(low)
+        if low.dimensionless:
+            low = Quantity(low, self._units)
         high = Quantity(high)
+        if high.dimensionless:
+            high = Quantity(high, self._units)
         TaurusAttribute.setRange(self, [low, high])
         infoex = self._pytango_attrinfoex
         if low.magnitude != float('-inf'):
@@ -829,7 +833,11 @@ class TangoAttribute(TaurusAttribute):
             limits = limits[0]
         low, high = limits
         low = Quantity(low)
+        if low.dimensionless:
+            low = Quantity(low, self._units)
         high = Quantity(high)
+        if high.dimensionless:
+            high = Quantity(high, self._units)
         TaurusAttribute.setWarnings(self, [low, high])
         infoex = self._pytango_attrinfoex
         if low.magnitude != float('-inf'):
@@ -847,7 +855,11 @@ class TangoAttribute(TaurusAttribute):
             limits = limits[0]
         low, high = limits
         low = Quantity(low)
+        if low.dimensionless:
+            low = Quantity(low, self._units)
         high = Quantity(high)
+        if high.dimensionless:
+            high = Quantity(high, self._units)
         TaurusAttribute.setAlarms(self, [low, high])
         infoex = self._pytango_attrinfoex
         if low.magnitude != float('-inf'):
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -812,10 +812,10 @@ class TangoAttribute(TaurusAttribute):
         low, high = limits
         low = Quantity(low)
         if low.dimensionless:
-            low = Quantity(low, self._units)
+            low = Quantity(low.magnitude, self._units)
         high = Quantity(high)
         if high.dimensionless:
-            high = Quantity(high, self._units)
+            high = Quantity(high.magnitude, self._units)
         TaurusAttribute.setRange(self, [low, high])
         infoex = self._pytango_attrinfoex
         if low.magnitude != float('-inf'):
@@ -834,10 +834,10 @@ class TangoAttribute(TaurusAttribute):
         low, high = limits
         low = Quantity(low)
         if low.dimensionless:
-            low = Quantity(low, self._units)
+            low = Quantity(low.magnitude, self._units)
         high = Quantity(high)
         if high.dimensionless:
-            high = Quantity(high, self._units)
+            high = Quantity(high.magnitude, self._units)
         TaurusAttribute.setWarnings(self, [low, high])
         infoex = self._pytango_attrinfoex
         if low.magnitude != float('-inf'):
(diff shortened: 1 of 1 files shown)


lib/taurus/core/tango/tangoattribute.py
```diff
@@ -811,7 +811,11 @@ def setRange(self, *limits):
             limits = limits[0]
         low, high = limits
         low = Quantity(low)
+        if low.dimensionless:
+            low = Quantity(low, self._units)
```
cpascual: This is breaking the test. 
Maybe a more correct implementation is:

```python
if low.dimensionless:
    low = Quantity(low.magnitude, self._units)
```

cpascual:
This implementation is buggy (see the broken tests). See a possible solution (applicable to the other methods as well) here: https://github.com/taurus-org/taurus/pull/395/commits/df0a8cf632c38c3f6b4f5e34066a0a133cf31324#r99134561

cpascual:
Also, I think that setRange, etc should issue a deprecation warning if they need to convert "implicit" values to quantites. Therefore, the logic should be a bit different:

```python
low = Quantity(low)
if low.dimensionality != self._units.dimensionality:
    low = Quantity(low.magnitude, self._units)
    self.deprecation(dep='calling TangoAttribute.setRange with implicit units', 
                               alt='explicit units', rel='4.0')
```

lib/taurus/core/tango/tangoattribute.py
cmft: I saw the travis report, but it is very rare, I could not reproduce the failures locally.

```
 python  -m unittest taurus.core.tango.test.test_tangoattribute.AttributeTestCase
MainThread     INFO     2017-02-02 16:08:45,569 Starter: Starting server TangoSchemeTest/unittest...
MainThread     INFO     2017-02-02 16:08:46,578 Starter: Server TangoSchemeTest/unittest has been started
MainThread     WARNING  2017-02-02 16:08:46,713 TaurusRootLogger: /home/cfalcon/tools/taurus.git/lib/taurus/core/tango/test/test_tangoattribute.py:814: DeprecationWarning: _get_w_value is deprecated since 4.0. Use .wvalue instead
  got = getattr(read_value, k)

MainThread     WARNING  2017-02-02 16:08:46,715 TaurusRootLogger: /home/cfalcon/tools/taurus.git/lib/taurus/core/tango/test/test_tangoattribute.py:814: DeprecationWarning: _get_value is deprecated since 4.0. Use .rvalue instead
  got = getattr(read_value, k)

..ss...............................MainThread     WARNING  2017-02-02 16:08:48,398 TaurusRootLogger: /home/cfalcon/tools/taurus.git/lib/taurus/core/tango/test/test_tangoattribute.py:802: DeprecationWarning: getUnit is deprecated since 4.0. Use .rvalue.units instead
  got = getattr(a, k)

MainThread     WARNING  2017-02-02 16:08:48,399 TaurusRootLogger: /home/cfalcon/tools/taurus.git/lib/taurus/core/tango/test/test_tangoattribute.py:814: DeprecationWarning: TangoAttrValue.type is deprecated since 4.0. Use TangoAttribute.type instead
  got = getattr(read_value, k)

MainThread     WARNING  2017-02-02 16:08:48,399 TaurusRootLogger: /home/cfalcon/tools/taurus.git/lib/taurus/core/tango/test/test_tangoattribute.py:814: DeprecationWarning: TangoAttrValue.unit is deprecated since 4.0. Use TangoAttribute.unit instead
  got = getattr(read_value, k)

.....................MainThread     INFO     2017-02-02 16:08:49,281 Starter: Hard killing server TangoSchemeTest/unittest...
MainThread     INFO     2017-02-02 16:08:49,288 Starter: Server TangoSchemeTest/unittest has been stopped
MainThread     INFO     2017-02-02 16:08:49,296 Starter: Deleted device TangoSchemeTest/unittest/temp-1
MainThread     INFO     2017-02-02 16:08:49,296 Starter: Hard killing server TangoSchemeTest/unittest...
MainThread     INFO     2017-02-02 16:08:49,296 Starter: Server TangoSchemeTest/unittest has been stopped
MainThread     INFO     2017-02-02 16:08:49,315 Starter: Deleted Server TangoSchemeTest/unittest

Ran 56 tests in 3.801s

OK (skipped=2)

```
About the implementation, I agree that maybe is better to use "low.magnitude", but with my Pint  version (0.8.dev0) it is valid. I will try with the docker.

```
from taurus.external.pint import Q_
q = Q_(1)
q2 = Q_(q, 'mm')
 print q2 #1 mm

```

cmft: >  I will try with the docker.

A priori it works, but when I ask for the version pint returns "'unknown-taurus'"

cpascual: > About the implementation, I agree that maybe is better to use "low.magnitude", but with my Pint version (0.8.dev0) it is valid. I will try with the docker.

Careful: pint accepts it, but it is not doing what you think. See this similar example:


```python
In [20]: q=Q_("1V")
In [21]: q2=Q_(q, "m")
In [22]: q2
Out[22]: <Quantity(1 V, 'meter')>
In [23]: print q2
1 V m

In [24]: q2.units
Out[24]: <Unit('meter')>
In [25]: q2.magnitude
Out[25]: <Quantity(1, 'volt')>
```
i.e. if you do not use `.magnitude`, you are creating a quantity whose magnitude is a *quantity*.

I think this does not make sense (definitely not in our case), but pint is quite liberal in what it accepts as a magnitude (in order to allow you to use arbitrary objects such as numpy arrays, etc) as the underlying magnitude in quantities.

cmft: Ups!! You are right!! I will fix it.

cmft:
> Also, I think that setRange, etc should issue a deprecation warning if they need to convert "implicit" values to quantites. Therefore, the logic should be a bit different:

I don't know, we don't do it in the "write" method, See:
```
import taurus
from taurus.external.pint import Q_

attr = taurus.Attribute('sys/tg_test/1/float_scalar')
attr._units = 'mm'
attr.write(1)
attr.write(Q_('1m'))
```
At the end, accept dimensionless quantities is a side effect of accept string arguments.

cpascual:
> we don't do it in the "write" method
You are right. We decided on doing it that way to avoid too many warnings and also because the implementation would be complex (not all attributes hold quantities as wvalue). But still I think that we should consider issuing the warnings for the setRange, etc methods, which **always** are related to quantities.

cpascual:
While integrating it, I just realized that the check should not be done on `.dimensionless` but on `.unitless`

PR #396, opened 2017-02-06
Fix bug#255 AttributeError in taurusplot

cmft:
Taurusplot has "tango-centric" code that raises AttributeError
exceptions when the default scheme is not "tango" 
(See #255).

Fix it, checking the default_scheme when the model is not set
avoiding to create a Tango widget (TaurusModelChooser).

### DIFF
--- a/lib/taurus/qt/qtgui/plot/taurusplot.py
+++ b/lib/taurus/qt/qtgui/plot/taurusplot.py
@@ -37,6 +37,7 @@ import numpy
 from taurus.external.qt import Qt, Qwt5
 
 import taurus
+from taurus import tauruscustomsettings
 import taurus.core
 from taurus.core.taurusmanager import getSchemeFromName
 from taurus.core.taurusbasetypes import DataFormat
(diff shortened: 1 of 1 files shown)


cpascual:
I do not think this is the best way to deal with it.  

Note that your fix does not prevent the same error occurring if you trigger the `_inputDataAction` (e.g. with `CTRL+N`), which also calls `TaurusPlot.showDataImportDlg()`. IMHO, the problem should be fixed at the `TaurusPlot.showDataImportDlg()` level or even better, at the `TaurusModelChooser` level.

Keep in mind for any implementation proposal, that the TaurusModelChooser should eventually become really scheme-agnostic and provide an entry-point for scheme plugins to insert their own specific model choosers. This is described in appendix IV of the [TEP13](https://github.com/taurus-org/taurus/blob/develop/doc/source/tep/TEP13.md).

cmft:
You are right, I did not  see the Ctrl+N option.

 I did not do it at TaurusPlot.showDataImportDlg() level keeping in mind that you commented "TaurusModelChooser should eventually become really scheme-agnostic " and it is a workaround. 

But it is better protect the code to this level plus a "TODO"

cpascual:
Hi, @cmft, If I understand correctly, we agree on fixing this at  `TaurusPlot.showDataImportDlg()`  or  `TaurusModelChooser` level. Should I reject this PR or should we reuse it?

cmft:
I did a new [PR](https://github.com/taurus-org/taurus/pull/428). I close it.

PR #403, opened 2017-02-22
Use unitless instead of dimensionless when handling implicit units

cpascual:
Several parts of taurus handle unitless quantities but check
using the .dimensionless attribute, which is not exactly the same:
e.g., Q_('1 rad') is dimensionless, but not unitless

Fix it by using the .unitless attribute instead.

### DIFF
--- a/lib/taurus/core/tango/util/tango_taurus.py
+++ b/lib/taurus/core/tango/util/tango_taurus.py
@@ -121,7 +121,7 @@ def unit_from_tango(unit):
     except (UndefinedUnitError, UnicodeDecodeError):
         # TODO: Maybe we could dynamically register the unit in the UR
         from taurus import warning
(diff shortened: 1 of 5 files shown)


teresanunez:
@cpascual, only to be sure. In the file external/pint/pint_local/unit.py line 115 is correct to put
 dimensionless ?.

cpascual:
@teresanunez 
> In the file external/pint/pint_local/unit.py line 115 is correct to put dimensionless ?

Not sure what you mean...
I do not see "dimensionless" in that line (I am using the develop branch... I see it in line 104 and line 123)

In any case, that file belongs to pint and we should not care about pint internals.

cpascual:
@cmft already approved, 
@teresanunez , I understand that you were also ok with merging, ... 

so I am merging

PR #412, opened 2017-02-24
Add API_PollObjNotFound as EVENT_TO_POLLING_EXCEPTIONS

sergirubio:
Hi,

When disabling server polling on Tango device servers, Taurus is not switching to client polling automatically.

We've seen that the received exception (API_PollObjNotFound) was not included in the EVENT_TO_POLLING_EXCEPTIONS enum.

Just adding it, the trends are capable of switching to polling automatically.

S.Rubio and M.Rosanes


sergirubio:
This fix only applies for Tango9  servers and clients. Tango7 switch to polling doesn't work with or without this fix.

cpascual:
Fix committed to develop after amending the author (@srubio) and the commit message (which was a bit uninformative).

sergirubio:
Hi @cpascual , be careful with the author @srubio as he is not me!!, I'm @sergirubio on github

PR #415, opened 2017-03-08
Add parentObj property to TaurusModel

cpascual:
Add a parentObj property with getPArentObj as getter to all TaurusModels.
This is useful in connection with fragments, since it allows one to refer to,
 e.g. "tango:a/b/c/d#parentObj.name"  to make a TaurusLabel show "a/b/c"

### DIFF
--- a/lib/taurus/core/taurusmodel.py
+++ b/lib/taurus/core/taurusmodel.py
@@ -300,3 +300,5 @@ class TaurusModel(Logger):
     @property
     def fullname(self):
         return self._full_name
+
+    parentObj = property(fget=getParentObj)


cpascual:
On a second thought... I have some doubts regarding the actual name of the property:

- I initially thought of using `parent` because it is simpler and nicer for using in fragments, but I discarded it because it is too likely to collide in subclasses (and there are too many different parentship relationships (class inheritance, Model parenting, Qt parenting,...) 
- In my proposal I chose `parentObj` because it relates to the `TaurusModel.getParentObj()`.
- But now I think that maybe a more explicitly model-related name such as `parent_model` or `model_parent` could be better (at the cost of not relating to the current API).

What do you think?

jairomoldes:
Trying to test a TaurusLabel setting the model name to:
tango://pc202:10000/bo/pc/kiext/state#parent.name

only shows "-----" instead of the device name

cmft:
I reviewed the enhancement, and I agree to merge it into develop.

PR #419, opened 2017-03-16
Fix taurusgui main

cmft:
The tauruslabel example set to lowercase the model.
This breaks the case-sensitive URIs.

Avoid to do it.


cpascual:
Manually merged to fix wrong commit message. The changes are already in develop.
Thanks!

PR #420, opened 2017-03-16
tangoattribute unitless warning

srgblnch:
The log message about the tango attributes unitless didn't say who is the raiser of this warning.

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -57,7 +57,7 @@ from .util.tango_taurus import (description_from_tango,
                                 display_level_from_tango,
                                 quality_from_tango,
                                 standard_display_format_from_tango,
-                                unit_from_tango, quantity_from_tango_str,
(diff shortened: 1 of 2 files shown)


cpascual:
Hi @srgblnch , thanks for the contribution. 
However, I have some doubts about the proposed implementation:

- I would use the full name instead of the simple name (consider e.g. gui controlling several motors, a warning saying that the problem is in "position" would not be very informative)
- I would avoid having to pass the name. 

IMHO, a better approach (that solves both issues) would be to move the code from `tango_taurus.unit_from_tango` to  the `TaurusAttribute` class, where `self.warning` could be used (with the original message) to convey all the info.

srgblnch:
I've added a commit to this pull request with a modification thinking in the suggested approach.

I think this is not the final one because I've seen this method 'unit_from_tango(unit)' in "tango_taurus.py" file is used in "test/test_tangoattribute.py" and I'm not seeing how I can modify the test to use the method with in the 'TangoAttribute' class.

PR #422, opened 2017-03-21
Fix bug #421 TarurusTrend does no work with boolean_scalar

### LINKED ISSUE #421: TarurusTrend does no work with boolean_scalar
taurustrend can not plot the values cause the events are been dropped, due to its rvalue has not .magnitude (it is due to booleans not being quantities)

> TaurusTrendsSet.sys/tg_test/1/boolean_scalar: Droping event. Reason rvalue has no .magnitude

cmft:
Taurus plotting widgets can not plot boolean attributes since
API changes in Taurus 4. They expected numeric values (Quantities).

Fix it, allowing also boolean attributes.

Fixes #421


cpascual:
I merged it, with a couple of changes of my own:

- Avoid calling isNumeric from a value (deprecated)
- fix typo

PR #423, opened 2017-03-22
Extend eval scheme to allow usage of arbitrary modules

cpascual:
A little known (yet very powerful) feature of the eval scheme is that you can use custom evaluator devices.

The problem with the existing solution is that it forces you to create an ad-hoc class for the custom evaluator. This PR simplifies the creation of custom evaluators. In this new implementation, one can use the syntax `eval:@foo.*` to indicate that an evaluator device should be created which has all symbols from the `foo` module available for the evaluation.

For example, one could do a simple clock with:

```
taurusform 'eval:@datetime.*/datetime.now().isoformat()'
```

Notes:

- This facilitates enormously the access to many sources of data without having to create an specific scheme.
- one big limitation of this is that the attributes are read-only (you would need a proper dedicated scheme for writing)

### DIFF
--- a/doc/source/devel/core_tutorial.rst
+++ b/doc/source/devel/core_tutorial.rst
@@ -98,9 +98,14 @@ Or one that adds noise to a tango image attribute:
     
 ``eval:img={tango:sys/tg_test/1/short_image_ro};img+10*rand(*img.shape)``
 
+And, by using custom evaluators, one can easily access virtually anything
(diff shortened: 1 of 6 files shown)


cpascual:
I am merging it since it got possitive reviews from @teresanunez and @cmft

cpascual:
Note. See PR #425, which addresses the main limitations left after this one.

PR #425, opened 2017-03-29
Add write and instantiation support to eval scheme

cpascual:
This PR extends the evaluation scheme with two related features (it is a continuation of PR #423 ):

- support for creation of arbitrary instances by EvaluationDevices
- support for writeable attributes

The first feature introduces the possibility of defining an Evaluation Device such as:

`eval:@c=mymod.MyClass(1,"a",b=3)`

This adds the `c` symbol to the context of the EvaluationAttributes of this device. In other words, 
any evaluation done by this Evaluation device is done in an environment that behaves as if 
one had done the following beforehand:

```python
from mymod import MyClass
c = MyClass(1, "a", b=3)
```

The second feature (support of writeable attributes) can be accessed by instantiating a class
that declares writeable properties. For example, assume that the `mymod` module defines the following class:

```python
from taurus.external.pint import Quantity


class MyClass(object):
    def __init__(self, foomag=123):
        self._foo = Quantity(foomag, "m")

    def get_foo(self):
        return self._foo

    def set_foo(self, value):
        self._foo = value

    foo = property(get_foo, set_foo)
```

Then, the following code would give a taurus form allowing to read and write foo (of course, a more practical example could do other calls in `get_foo` and `set_foo`):

`taurusform 'eval:@c=mymod.MyClass(99)/c.foo'`

For more examples, see the docs in the [evaluation module](https://github.com/taurus-org/taurus/blob/6d9b28f85d3e9da07887eb97908305a09d52e359/lib/taurus/core/evaluation/__init__.py) and in the [evaluation test resource files](https://github.com/taurus-org/taurus/tree/6d9b28f85d3e9da07887eb97908305a09d52e359/lib/taurus/core/evaluation/test/res)

Other possible ideas for using this:

- Getting images from a webcam without tango (with opencv)
- Make a trend of a log file size (with os.path.getsize)
- Read/write environment variables from a TaurusForm (with os.environ)
- ...


cpascual:
I added an example of how to use the new features to control an IcePAP in:
https://gist.github.com/cpascual/c1e1c1b8883e373efdd642663d23abd1

cpascual:
Hi @teresa:

>  you forgot to close the argument of taurusform with '' 

Fixed, thanks!

>  I have tried to use it for writing an environment variable, but if I do: 
 `taurusform 'eval:@os.*/environ["PYTHONPATH"]'` 
I get a taurusfom showing the PYTHONPATH value but it doesn't allow to write it 

In order to have writeable attributes you need to have a class declaring properties (because eval needs to know both the getter and the setter). You can see an example of a class that uses `os.environ` for read/write attributes in the [mymod.py example](https://github.com/cpascual/taurus/blob/0327ba2c2b6f588ed99708b9fcf14b33e53c52b5/lib/taurus/core/evaluation/test/res/mymod.py) (see the `MyClass.bar` property)

cpascual:
@teresa, since you are already testing it, do not hesitate to integrate it yourself and merge if you think it is ok

PR #427, opened 2017-03-31
Make TaurusValuesTable show booleans as checkboxes

cpascual:
TaurusValuesTable shows booleans as "True" or "False" strings. Use checkboxes instead

To reproduce it, use 
```
taurusform sys/tg_test/1/boolean_spectrum
```

### DIFF
--- a/lib/taurus/qt/qtgui/table/taurusvaluestable.py
+++ b/lib/taurus/qt/qtgui/table/taurusvaluestable.py
@@ -101,7 +101,25 @@ class TaurusValuesIOTableModel(Qt.QAbstractTableModel):
             tabledata = self._wtabledata
         if not index.isValid() or not (0 <= index.row() < len(tabledata)):
             return Qt.QVariant()
-        elif role == Qt.Qt.DisplayRole:
+        elif (role == Qt.Qt.CheckStateRole and
+                self.typeCastingMap[tabledata.dtype.kind] == bool):
+            value = None
+            rc = (index.row(), index.column())
+            if self._writeMode and rc in self._modifiedDict:
+                value = self._modifiedDict[rc]
+            else:
+                value = tabledata[rc]
+                if isinstance(value, Quantity):
+                    value = value.magnitude
+            # cast the value to a standard python type
+            value = self.typeCastingMap[tabledata.dtype.kind](value)
+            if value:
+                value = Qt.Qt.Checked
+            else:
+                value = Qt.Qt.Unchecked
+            return value
+        elif (role == Qt.Qt.DisplayRole and
+                self.typeCastingMap[tabledata.dtype.kind] != bool):
             value = None
             rc = (index.row(), index.column())
             if self._writeMode and rc in self._modifiedDict:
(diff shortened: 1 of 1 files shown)


cpascual:
Hi @jairomoldes , I've reviewed the proposed change and I must say that **personally** I prefer the previous behaviour for the following reasons:

- In the case of writeable attributes, if you enter the edition mode (double-click on a cell, or select a cell and press F2), you see **2** checkboxes side-to-side: one coming from the `CheckStateRole` and another coming from the `EditRole`. Only the second can be used for changing the state. IMHO this is more confusing than the current combobox.
- IMHO, the checkbox alone (with no text) is a bit confusing regarding its meaning (I find the string "true" and "false" more explicit.
- *When the cell is selected*, checked checkbox are not easy to distinguish from unchecked ones (at least with my desktop theme)

**But** I am aware that this may be just a matter of taste, so before taking a decision on merging/rejecting the PR, I would like hear other opinions ( @taurus-org/integrators , @cmft , @sagiss , ... anyone reading this... what, do you think?)

cmft:
After test the changes, I agree with the @cpascual's arguments. 
I also prefer the combo-box, so I would not merge this PR.

teresanunez:
I have also test this and I agree with @cpascual and @cmft, I prefer the combo-box.

jairomoldes:
Hi @cpascual. About you 3 comments:
- About the 2 checkboxes. Actually this is the easiest implementation, and I think that the user can see what was the value before he checks/unchecks the EditRole.
- IMHO displaying a large array of booleans with True/False text makes it hard to actually distinguish them. About the meaning of True/False I think that it is quite a standard interpretation that checked is True and unchecked is False.
- This doesn't happen in my desktop. It is quite easy to distinguish even if they are selected.

jairomoldes:
I would like to illustrate what I would like to get with pull requests:
https://github.com/taurus-org/taurus/pull/443
https://github.com/taurus-org/taurus/pull/427

I have an 88 boolean spectrum attribute that represent the 88 BPMs that can be used for our FOFB systems. These 88 BPMs are distributed in the 16 sectors of the machine and furthermore the number of BPMs per sector may be different. I think that this feature would be of general use for other people.

Have a look at the attached images to see the difference between current status and what I propose with my PRs.

boolarray1
boolarray2

cpascual:
Just thinking out loud...

If I understand correctly, the motivation for this is to facilitate a more compact display and to make it easier to see the different states, so, wouldn't it be better if we used background color (e.g. green=True & gray=False) without any text but with a tooltip (compact) and then a *single* checkbox when we are in edition mode?

PS: in my proposed colors I am avoiding red because of its "fault" meaning in Taurus. Also, the "green" that I would choose would be different (darker) than the one we use for the OK quality

@jairomoldes  what do you think?

mrosanes:
After testing both options, in debian9 I also face the problem 3 of @cpascual (when cell is selected...)

I would like the option of changing the background color (ex: green for True, red for False: using darker colors in order to not confuse them with Taurus colors). It would be nice to have the option to change the background color (from True to False) with only one click of the mouse.

Also it would be interesting if a single letter T or F (standing for True and False), could be added on top of the background color.

However, if I have to chose between the current options, I prefer the combobox with True/False.

cpascual:
@jairomoldes : after some debate we decided against merging this PR (in its current form).
We agree in that usability/readibility could be improved in the table widget but we think that the issues introduced in this implementation need to be addressed before merging.

If you are interested in proposing a new implementation that addresses them, please create a new PR. Alternatively, if you do not have the time to do it yourself but think it is worth to record the request, please open an issue.

PR #428, opened 2017-03-31
Fix Attribute errors when use eval as default scheme

cmft:
TaurusModelChooser is still Tango-Centric, but it is showed when a model is not given in taurusplot, or taurusform.

Avoid to launch it when tango is not the default scheme.

Fix bug #255.

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusform.py
+++ b/lib/taurus/qt/qtgui/panel/taurusform.py
@@ -36,6 +36,7 @@ import PyTango
 
 import taurus.core
 from taurus.core import TaurusDevState
+from taurus import tauruscustomsettings
 
 from taurus.qt.qtcore.mimetypes import (TAURUS_ATTR_MIME_TYPE, TAURUS_DEV_MIME_TYPE,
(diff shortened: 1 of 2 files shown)


cpascual:
Hi, thanks for the contribution, but I do not think this is the right approach (sorry!)
The workaround should be done in TaurusModelChooser, not in TaurusPlot/ TaurusForm, etc.

The reasons:

- The proposed implementation leaves tango-specific code that will need to be re-found and removed later on
- The proposed implementation disables other ways of adding curves (non-models) to TaurusPlot
- The TaurusModelChooser is Tango-centric, but not completely so. The upper part (the tree) is 100% tango, but the bottom part (the list) is, in principle, scheme agnostic. The list is currently the only way for a user to edit the models (also non-tango models). If you prevent the ModelChooser from loading, you are also cripling this access.
- The default scheme should *not* be used as an indicator of wether Tango is available or not. One may have Tango installed and not make it the default scheme.

So, IMHO, a better approach (short of actually making TaurusModelChooser really agnostic) is to add some workaround allowing the TaurusModelChooser to launch even without Tango (and maybe disabling the Tango tree also).

PR #430, opened 2017-04-04
(doc) Update contrib instructions in devel guide

cpascual:
Coding guide still points to old contribution policy. Update it.


cpascual:
@taurus-org/integrators This is a trivial patch. Please integrate

PR #444, opened 2017-04-20
Fix  Bug 392: Wrong formatting of Tango attributes

cmft:
According the discussion in the ticket #392:
- We add a "precision" attribute to the TaurusAttribute class with the same meaning as in pyEpics ("number of decimal places of precision for floating point values")
- We encode/decode it for each scheme (e.g. for tango, we would have: format="%6.2f" <--> precision=2 )
- We make the display widgets use this precision as a hint for display
- We add an API to set a formatter in TaurusBaseComponent (allowing python str format or callable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ develop branch)  won't be reflected in this file.
 
 ## Unreleased
 
+### Added
+- Formatting API in TaurusBaseComponent (#444)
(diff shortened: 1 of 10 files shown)


teresanunez:
I agree with @cpascual, I see it ready to be merged.

LJBD:
Looks all right to me and fulfils what I've expected when creating #392 .

Issue #398, opened 2017-02-07
taurus.qt.qtgui fails to import if PyTango is not installed


cpascual:
To reproduce, **uninstall PyTango** and do:

```
:~$ python -c 'import taurus.qt.qtgui'
MainThread     INFO     2017-02-07 09:50:03,482 TaurusRootLogger: Using "PyQt4" for Qt
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/__init__.py", line 35, in <module>
    import icon as __icon
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/icon/__init__.py", line 32, in <module>
    from catalog import QIconCatalog
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/icon/catalog.py", line 31, in <module>
    from taurus.qt.qtgui.input import GraphicalChoiceWidget
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/input/__init__.py", line 32, in <module>
    from .tauruscheckbox import *
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/input/tauruscheckbox.py", line 33, in <module>
    from taurus.qt.qtgui.base import TaurusBaseWritableWidget
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/base/__init__.py", line 31, in <module>
    from .taurusbase import *
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/base/taurusbase.py", line 56, in <module>
    from taurus.qt.qtgui.util import ActionFactory
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/util/__init__.py", line 33, in <module>
    from .tauruscolor import *
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/util/tauruscolor.py", line 40, in <module>
    from taurus.core.tango import DevState  # TODO: Tango-centric
  File "/nfs/home/cpascual/src/taurus/lib/taurus/core/tango/__init__.py", line 108, in <module>
    from enums import *
  File "/nfs/home/cpascual/src/taurus/lib/taurus/core/tango/enums.py", line 43, in <module>
    import PyTango
  File "/usr/lib/python2.7/dist-packages/PyTango.py", line 5, in <module>
    import tango
[...]

PR #446, opened 2017-05-05
Fix bug #398 taurus.qt.qtgui fails to import

cmft:
taurus.qt.qtgui fails to import if PyTango is not installed.

Since taurus.qt.qtgui core is still not Tango independent,
as a workarround the exception is catching to avoid the ImportError

Fixes: #398


cpascual:
Note that the travis failure seems a false positive (probably related to #261).
Running the failing test manually returns all ok:
```
python -m unittest taurus.qt.qtgui.display.test.test_tauruslabel.TaurusLabelTest2
(...)
Ran 13 tests in 14.848s
OK
```

cpascual:
I am merging this PR because it does solve the import problem for taurus.qt

But note that the problem is more complex and needs a full revision of taurus.qt to make at least all the scheme-agnostic widgets importable without PyTango installed.

Issue #433, opened 2017-04-05
Unexpected rvalue type in eval attribute with quantities


cmft:
The EvalAttribute when the URI explicit defines a Quantity, e.g. "eval:Q('1m')", the  rvalue type returns a `numpy.ndarray` instead of scalar (int/float), although the attribute data_format says it is a scalar.

You can reproduce it with this code:
```
import taurus
a = taurus.Attribute("eval:Q('1m')")
a.data_format
Out[14]: 0 # SCALAR
type(a.read().rvalue.magnitude)
Out[15]: numpy.ndarray
```

cmft:
The  bug is in `evalAttribute applyTransformation` method.  The uses of `numpy.shape` method over a Quantity mutate the Quantity's magnitude type.

```
In [1]: import numpy

In [2]: from taurus.external.pint import Quantity

In [3]: v = Quantity(1)

In [4]: print type(v.magnitude)
<type 'int'>

In [5]: value_dimension = len(numpy.shape(v))

In [6]: print type(v.magnitude)
<type 'numpy.ndarray'>
```

cmft:
I opened an issue in pint [#509](https://github.com/hgrecco/pint/issues/509)

cpascual:
fixed in #448

PR #448, opened 2017-05-16
Fix Bug #433 Unexpected rvalue type in eval attribute (with quantities)

cmft:
The evalAttribute rvalue type returns a numpy.ndarray instead of
scalar (int/float) when the URI explicit defines a Quantity.

e.g. "eval:Q('1m')"

Apply numpy.shape method over a Quantity mutates the
magnitude type.

### DIFF
--- a/lib/taurus/core/evaluation/evalattribute.py
+++ b/lib/taurus/core/evaluation/evalattribute.py
@@ -334,7 +334,10 @@ class EvaluationAttribute(TaurusAttribute):
         try:
             evaluator = self.getParentObj()
             rvalue = evaluator.eval(self._transformation)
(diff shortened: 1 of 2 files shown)


cpascual:
Since this seems to be caused by a bug in pint, it would be nice to add some comment and a TODO marking this as a workaround to be removed when/if pint fixes it

cmft:
> Since this seems to be caused by a bug in pint, it would be nice to add some comment and a TODO marking this as a workaround to be removed when/if pint fixes it

Done!!

cpascual:
@cmft reported the issue in pint: https://github.com/hgrecco/pint/issues/509

PR #426, opened 2017-03-30
fix _info due to Unpacking error

piertoni:
On Windows platform causes unpacking error as an additional parameter is delivered, in this case is 'Starter', as I don't know the rest of the code I have assigned a temporary name that should be changed with the appropriated name.
I found this error using taurusgui --new-gui on Model Choser, without this was causing unpacking error, with this additional parameter it seems to work

PR #458, opened 2017-05-25
Fix error in TangoDatabaseCache (for recent Tango)

cpascual:
TangoDatabaseCache.refresh gets an unpacking error when 
connecting to a Tango 9.2 Database because the DbGetDeviceInfo 
command changed its signature. Fix it to support both old and 
new signature.

See closed PR #426 for more information.


cpascual:
Note that still there is the open question on why is this code accessed (since it should only be reached as a fall-back for non-mysql DBs). 

@sergirubio was looking into it

cpascual:
@sergirubio : any news on the investigation of why the new database DS makes taurus to use the unoptimized code?

cpascual:
@teresanunez Would you mind integrating this? (it should be safe to do it as-is, we will take care of changing the unoptimized issue later on...)

PR #466, opened 2017-06-15
Ignore harmless exception in AutoScrollTool

cpascual:
A harmless exception is thrown every time that the AutoScrollTool is
loaded. It is related to an attempt to disconnect a not connected
signal. Handle the exception to ignore it.


cpascual:
@taurus-org/integrators this is ready for being reviewed

PR #469, opened 2017-06-16
Needed for sardana issue 102. Adding loggingtofile to logger

teresanunez:
The implemention of a new logging for sardana (sardana-org/sardana#102) requires this change in taurus. A new method (loggingtofile) is added to the Logger.

### DIFF
--- a/lib/taurus/core/util/log.py
+++ b/lib/taurus/core/util/log.py
@@ -793,7 +793,7 @@ class Logger(Object):
                         out += '<could not find suitable string representation>'
         return out
 
-    def loggingtofile(self, msg, logging_path, *args, **kw):
+    def log_to_file(self, msg, logging_path, *args, **kw):
 
         file_name = logging_path + "/spock_session.log"
 
--- a/lib/taurus/core/util/log.py
+++ b/lib/taurus/core/util/log.py
@@ -797,10 +797,10 @@ class Logger(Object):
 
         file_name = logging_path + "/spock_session.log"
 
-        logging_file = open(file_name, 'a')
-        logging_file.write('\n')
-        logging_file.write(msg)
-        logging_file.close()
+        with open(file_name, 'a') as logging_file:
+            logging_file.write('\n')
+            logging_file.write(msg)
+            logging_file.close()
 
     def log(self, level, msg, *args, **kw):
         """Record a log message in this object's logger. Accepted *args* and
--- a/lib/taurus/core/util/log.py
+++ b/lib/taurus/core/util/log.py
@@ -793,6 +793,15 @@ class Logger(Object):
                         out += '<could not find suitable string representation>'
         return out
 
+    def loggingtofile(self, msg, logging_path, *args, **kw):
+
+        file_name = logging_path + "/spock_session.log"
+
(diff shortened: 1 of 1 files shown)


lib/taurus/core/util/log.py
```diff
@@ -793,6 +793,15 @@ def _format_stack(self, stack_func=inspect.stack):
                         out += '<could not find suitable string representation>'
         return out
 
+    def loggingtofile(self, msg, logging_path, *args, **kw):
```
LJBD: IMHO this method's name should be written with `snake_case` and begin with a verb. What do you think of: `log_to_file`?

teresanunez: Not problem in changing the name. I would have to change it after that in the Sardana PR.

lib/taurus/core/util/log.py
```diff
@@ -793,6 +793,15 @@ def _format_stack(self, stack_func=inspect.stack):
...
+
+        file_name = logging_path + "/spock_session.log"
```
cpascual: it would be neater if using os.path.join. But it is not critical

lib/taurus/core/util/log.py
```diff
@@ -793,6 +793,15 @@ def _format_stack(self, stack_func=inspect.stack):
...
+
+        logging_file = open(file_name, 'a')
+        logging_file.write('\n')
+        logging_file.write(msg)
+        logging_file.close()
+
```
cpascual: it would be nice to protect this against IO errors. For example, lines 800-803 can be better rewritten as:

```
with open(file_name, 'a') as logging_file:
    logging_file.write('\n')
    logging_file.write(msg)
```

cpascual: Now a more general comment. My knowledge of the `logging` module is a bit rusty, but I think that the feature that you are implementing with this `loggingtofile` method may already be covered in the `logging` module in a more generic form.
@cmft is the expert on this and may have something to say

teresanunez: Hi @cpascual, thanks, it was protected in the sardana part, informing about the reason of the
 error, but I put it now here and I took out the reason of the error (since now  could be another
 one) in the sardana part.

cmft: Without knowing the uses cases in Sardana. As @cpascual  pointed, the logging module allows you to redirect the output to a file or add a new (file) handler. See [1], and [2]

So, HIMO, I would not write the log with taurus.

Take in mind that Taurus.Logger provides the `getLogger (name)` method to get the appropriate Logger (e.g to access a subLogger),  so you can implement a solution based on  a `file handler` using the `logging` API (addHandler, removeHandler).

I will check the `log_to_file` uses and I'm coming with more ideas.

[1] https://docs.python.org/2/howto/logging.html#logging-to-a-file
[2] https://stackoverflow.com/questions/13733552/logger-configuration-to-log-to-file-and-print-to-stdout

teresanunez:
Hi, we are discussing about another way to implement the logging into a file in sardana. This new
 mode would not require changes in taurus. So this PR can be postponed until a decission about
 the sardana logging is taken.

PR #473, opened 2017-06-20
(Fix #401) Work-around bug introduced by qtpy

### LINKED ISSUE #401: qtpy exception (setResizeMode is only available in Qt4)
It seems that a deprecated Qt4 method is called:

  File "/home/sicilia/src/taurus/lib/taurus/qt/qtgui/table/taurusdevicepropertytable.py", line 186, in updateStyle
    QtGui.QHeaderView.ResizeToContents)  # .Stretch)
  File "/usr/lib/python2.7/dist-packages/qtpy/_patch/qheaderview.py", line 83, in setResizeMode
    raise Exception('setResizeMode is only available in Qt4. Use '
  Exception: setResizeMode is only available in Qt4. Use setSectionResizeMode instead.

cpascual:
Force an early import of qtpy and use pyqt5 API for QHeaderView in
order to work around bug https://github.com/taurus-org/taurus/issues/401

See also https://github.com/spyder-ide/qtpy/issues/119

Fixes #401

### DIFF
--- a/lib/taurus/external/qt/QtGui.py
+++ b/lib/taurus/external/qt/QtGui.py
@@ -33,5 +33,9 @@ _updateQtSubModule(globals(), "QtGui")
 
 if __backend == 'PyQt5':
     _updateQtSubModule(globals(), "QtWidgets")
+else:
+    # early import of qtpy.QtWidgets as a workaround for
+    # https://github.com/taurus-org/taurus/issues/401
+    import qtpy.QtWidgets as __qtpy_QtWidgets
 
 del _updateQtSubModule, getQtName
--- a/lib/taurus/qt/qtgui/table/qlogtable.py
+++ b/lib/taurus/qt/qtgui/table/qlogtable.py
@@ -489,7 +489,7 @@ class QLoggingWidget(QBaseTableWidget):
             klass = QLoggingTable
         view = QBaseTableWidget.createViewWidget(self, klass=klass)
         hh = view.horizontalHeader()
-        hh.setResizeMode(MSG, Qt.QHeaderView.Stretch)
+        hh.setSectionResizeMode(MSG, Qt.QHeaderView.Stretch)
         view.setShowGrid(False)
         view.sortByColumn(TIME, Qt.Qt.AscendingOrder)
         return view
--- a/lib/taurus/qt/qtgui/table/taurusdevicepropertytable.py
+++ b/lib/taurus/qt/qtgui/table/taurusdevicepropertytable.py
@@ -181,12 +181,12 @@ class TaurusPropTable(QtGui.QTableWidget, TaurusBaseWidget):
         headerItem1.setText(QtGui.QApplication.translate(
             "PLCTabWidget", "Value", None, QtGui.QApplication.UnicodeUTF8))
         self.setHorizontalHeaderItem(1, headerItem1)
-        self.horizontalHeader().setResizeMode(
+        self.horizontalHeader().setSectionResizeMode(
             QtGui.QHeaderView.ResizeToContents)  # .Stretch)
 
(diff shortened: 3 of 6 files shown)


cpascual:
Ready for review by @taurus-org/integrators

cmft:
Hi @cpascual ,
If  I do  a test with `taurusform` it raises an exception.  The back. comp. is broken with with the early import of qtpy. I think it should be protected,  and the user must be warned of this change.

In my opinion this is an important change (so, the changelog must be updated), and it should affect to the taurus version (minor number?)

Moreover the dependency of qtpy must be add.
 
```
taurusform sys/tg_test/1/boolean_spectrum
MainThread     INFO     2017-06-23 09:50:06,224 TaurusRootLogger: Using "PyQt4" for Qt
  File "/usr/bin/taurusform", line 9, in <module>
    load_entry_point('taurus==4.0.4-alpha', 'gui_scripts', 'taurusform')()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 356, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2476, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2190, in load
    ['__name__'])
  File "/home/cfalcon/tools/taurus.git/lib/taurus/qt/qtgui/__init__.py", line 35, in <module>
    import icon as __icon
  File "/home/cfalcon/tools/taurus.git/lib/taurus/qt/qtgui/icon/__init__.py", line 31, in <module>
    from icon import *
  File "/home/cfalcon/tools/taurus.git/lib/taurus/qt/qtgui/icon/icon.py", line 52, in <module>
  File "/home/cfalcon/tools/taurus.git/lib/taurus/external/qt/Qt.py", line 29, in <module>
    from .QtGui import *
  File "/home/cfalcon/tools/taurus.git/lib/taurus/external/qt/QtGui.py", line 39, in <module>
    import qtpy.QtWidgets as __qtpy_QtWidgets
ImportError: No module named qtpy.QtWidgets

```

cmft:
> Moreover the dependency of qtpy must be added.

#479 does it

cpascual:
> If I do a test with taurusform it raises an exception. The back. comp. is broken with with the early import of qtpy. 

bck-compat is not really broken. The old code works fine with the new version. What is true is that it forces a dependency being installed (which before was only relevant for a few widgets). Since this is dealt-with in #479, I propose to merge this PR only after #479 gets integrated

> I think it should be protected, and the user must be warned of this change.

I do not think it is a good idea. When #479 is merged, qtpy will be a dependency for taurus-qt, so there is no need to handle the case in which qtpy is not installed

> In my opinion this is an important change (so, the changelog must be updated),

I agree, but since we are on a bug-squashing spree, I opted for not updating the changelog on each PR to avoid handling merge conflicts. We will update the changelog for this (and other PRs in the Jul17 milestone in one go).

> and it should affect to the taurus version (minor number?)

Actually I think that this change is strictly a patch (no new features are added). 

In relation to this, note that currently we are not being strict in bumping the version on each merge to develop. Instead we relax this while we have the -alpha flag in the version and we do a full review of version and CHANGELOG prior to a release bump (this policy may change when/if we go deeper into Continuous Deployment).

As a side note, Jul17 release will surely imply a minor version bump (due to the many enhancements already introduced by other merges to develop).

reszelaz:
I agree with this solution. I decided to integrate this PR before https://github.com/taurus-org/taurus/pull/479 just because of the commodity. So for the moment the dependency of qtpy is not explicitly said. After integrating https://github.com/taurus-org/taurus/pull/479 it will be solved.

PR #477, opened 2017-06-22
make taurusgui --new-gui create a setup.py

### LINKED ISSUE #220: make taurusgui --new-gui create a setup.py
As part of the creation of a new gui with the wizard, a setup.py file could be created. Then instead of the message saying "put the folder in the PYTHONPATH" we could simply say: install it with "python setup.py" (or even offer to install right away)

cmft:
Refactor taurusgui wizard to generate a setup.py
based on setuptools.

Refactoring changes:
- Add python package structure.
- The setup.py, __init__.py, and MANIFEST.in file have been
created via templates.
- The launcher generated has been removed.

Fix #220


lib/taurus/qt/qtgui/taurusgui/taurusgui.py
```diff
@@ -25,7 +25,7 @@
 
 """This package provides the TaurusGui class"""
 
-__all__ = ["DockWidgetPanel", "TaurusGui"]
+__all__ = ["DockWidgetPanel", "TaurusGui", "main"]
```
cpascual: Is there any reason for exporting the "main" function?
In general we avoid it (we do not consider them public API), and entry points use a full path to access them

lib/taurus/qt/qtgui/taurusgui/res/setup.template
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+
+"""
+This file was autogenerated by taurusgui --new-gui.
+"""
+
+from setuptools import setup, find_packages
+
+# version (bumpversion style)
```
cpascual: maybe it would make sense to generate the .bumpversion file as well...

lib/taurus/qt/qtgui/taurusgui/res/setup.template
```diff
@@ -0,0 +1,37 @@
...
+    name="tgconf_{name}",
+    version=__version,
+    include_package_data=True,
+    packages=find_packages(),
+    entry_points=_entry_points,
+    url='http://www.taurus-scada.org/en/latest/',
+    keywords='APP',
+    description='{name} Taurus GUI',
+    requires=['setuptools (>=1.1)'],
+    install_requires=['taurus'],
```
cpascual: here we could use 'taurus[taurus-qt]' to be more precise about the dependencies (assuming that PR #479 is merged)

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
cmft: It is needed by the generated launcher (the method used by the entry point needs to access to this method)

lib/taurus/qt/qtgui/taurusgui/res/setup.template
cmft: I agree, we can change it when #479 has been integrated.

lib/taurus/qt/qtgui/taurusgui/res/setup.template
cmft: I don't know. For the one hand it does not disturb and for the other hand, I think is the developer who has to take the decision to use `bumpversion` and how to use it. 
Take in mind that the generated project are not under a git repo (Shall we do it?) 

I would not do more project configuration.

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
cpascual: Are you sure?
which generated launcher?
the `taurusgui` launcher already worked without this...

lib/taurus/qt/qtgui/taurusgui/res/setup.template
cpascual: > I think is the developer who has to take the decision to use bumpversion and how to use it.

The user could just ignore the .bumpversion file and work without it... but if we provide it we are encouraging/facilitating a good practice

>  generated project are not under a git repo

that is not relevant. bumpversion can also be used without git

> I would not do more project configuration.

That's fine by me. It was only a suggestion

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
cmft: > Which generated launcher?

It is done  for the gui launcher (for the method called in the entry point).
I did in the run method : 'from taurus.qt.qtgui.taurusgui import main' 

Checking  taurus 3 this can be done importing the main like this:
 `from taurus.qt.qtgui.taurusgui.taurusgui import main` 

I will change the code.

cpascual:
Tested and added 2 minor changes. All ok. Merging

PR #479, opened 2017-06-22
Update & refactor dependencies (Fix #376, #406, #407)

### LINKED ISSUE #376: Setuptools dependencies are outdated
From pypi (4.0.3) I see that taurus 4.0.3 declares the following deps:

guiqwt (==2.3.1),
spyder (>=2.2), 
lxml (>=2.1), 
ply (>=2.3), 
PyQt4.Qwt5 (>=5.2.0), 
PyQt4 (>=4.8), 
PyTango (>=7.1), 
numpy (>=1.1) 

Check the real current dependencies and update setup.py 
For example, spyder should be >=3 (or maybe not require it at all if we consider it optional)

### LINKED ISSUE #406: check_dependencies could be removed
Since the move of taurus to use setuptools, the check_dependencies mechanism (implemented in taurushelper.py) is kind of redundant and maintaining it is burdensome (dependencies need to be updated twice) . IMHO dependencies should be taken care of by the distribution method (setuptools, or linux distro, or whatever), so I propose to deprecate and remove check_dependencies() and related stuff

### LINKED ISSUE #407: pip does not install dependencies
Reported on Windows 10. See https://github.com/taurus-org/taurus/pull/397#issuecomment-281933818

cpascual:
- Update requirements (fixes #376)
- Use setuptool's "install_requires" instead of distutil's "requires"
  in order to fix #407 (pip does not install taurus dependencies)
- make most dependencies optional and associate each with an "extra"
  feature in Taurus.
- Refactor check_dependencies() (implemented in taurushelper.py) to
  make use of the dependencies declared in setup.py (fixes #406 )
- Change getting started documentation to facilitate maintenance
  of dependencies documentation.
- Deprecate log_dependencies()
- Clean unused private functions in `taurushelper.py`

### DIFF
--- a/doc/source/users/getting_started.rst
+++ b/doc/source/users/getting_started.rst
@@ -141,7 +141,7 @@ installation method:
 
       pip install taurus[NAME_OF_EXTRA]
 
-- The conda Conda_ package management system may also be used to install
+- The Conda_ package management system may also be used to install
   most of the required dependencies.
 
 - The `taurus-test Docker container`_ provides a Docker container (based
--- a/doc/source/users/getting_started.rst
+++ b/doc/source/users/getting_started.rst
@@ -22,8 +22,9 @@ You can test the installation by running::
 
        python -c "import taurus; print taurus.Release.version"
 
-
 Note: pip is already included in python>2.7.9
+
+Note: some "extra" features of taurus have additional dependencies_.
 
 Installing from sources manually (platform-independent)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -34,11 +35,13 @@ You may alternatively install from a downloaded release package:
 #. Extract the downloaded source into a temporary directory and change to it
 #. run::
 
-       python setup.py install
+       pip install .
 
 #. test the installation::
 
        python -c "import taurus; print taurus.Release.version"
+
+Note: some "extra" features of taurus have additional dependencies_.
 
 Linux (Debian-based)
 ~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


cpascual:
@taurus-org/integrators : this is ready for being reviewed.

teresanunez:
Hi @cpascual, I have been looking at this. IMO it could be merged.
 Only one stupid thing (not important at all, but for keeping the file cleaner), the even commented lines:

   # 'Taurus-Qt-TaurusGUI',  # [Taurus-TaurusGUI]
   # 'Taurus-Qt-Editor',  # [Taurus-Editor] --> or maybe move it to sardana

 should not be:

   # 'Taurus-Qt-TaurusGUI',  # [Taurus-Qt-TaurusGUI]
   # 'Taurus-Qt-Editor',  # [Taurus-Qt-Editor] --> or maybe move it to sardana

?

cpascual:
@teresanunez Thanks for reviewing!
I just changed the comments that you mentioned.

jkotan:
Hi,

When I try to run taurus-4.0.4 from this PR  I get the following error:

    jkotan@haso228k:~$ taurusgui macrogui  --taurus-log-level=debug 
      File "/home/jkotan/testinst/bin/taurusgui", line 9, in <module>
	load_entry_point('taurus==4.0.4-alpha', 'gui_scripts', 'taurusgui')()
      File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 356, in load_entry_point
	return get_distribution(dist).load_entry_point(group, name)
      File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2476, in load_entry_point
	return ep.load()
      File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2188, in load
	self.require(env, installer)
      File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2202, in require
	items = working_set.resolve(reqs, env, installer)
      File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 639, in resolve
	raise DistributionNotFound(req)
    pkg_resources.DistributionNotFound: qtpy

I hope Taurus still supports  the debian packaging system and setup.py and I don't have to use pip.

PS: develop of 4.0.4 does not work because of 

    ImportError: No module named qtpy.QtWidgets

Cheers,
Jan

jkotan:
Of course commenting qtpy from setup.py works.

cpascual:
Hi,

> When I try to run taurus-4.0.4 from this PR I get the following error:
> `jkotan@haso228k:~$ taurusgui macrogui  --taurus-log-level=debug `
> (...)
> `pkg_resources.DistributionNotFound: qtpy`

Do you have qtpy installed?
after this PR, qtpy  will be a dependency of Taurus 

So I only see a problem if you **do** have qtpy installed and taurus is failing to detect it...

> I hope Taurus still supports the debian packaging system and setup.py and I don't have to use pip.

Of course. We support both. But keep in mind that when using the git code you are not using the debian package (and therefore, if there are new depenedencies you may need to install them by hand). Once the new version of taurus gets packaged for debian, the new dependencies should be taken care of as usual.
 
> PS: develop of 4.0.4 does not work because of
> ImportError: No module named qtpy.QtWidgets

Same as before: make sure to have qtpy installed

reszelaz:
Reviewed and accepted. Thanks!

We should just keep in mind that the `guiqwt` project does not handle well its PyPI dependency `guidata`. Which must be installed manually before installing `taurus-qt`.

jkotan:
WRT qtpy: yes, you are right. Thanks. I forgot that for `jessie` we have `taurus 3.7.3` and `taurus4`is  for `stretch`.

PR #482, opened 2017-06-22
Allow subscribing to events w/ or w/o first event

reszelaz:
Emitting first event during the subscription process with the most recent
value may not be always desired.

This is the case of the TangoAttributeEG class in Sardana, a subclass of
the EventGenerator. It may work with the attribute which serves as a data
pipe for transferring scan results. It is not desired to receive an event
with the data corresponding to the previous scan while subscribing to events
of the new scan.

Allow control whether to emit the first event during the subscription process.

### DIFF
--- a/lib/taurus/core/util/event.py
+++ b/lib/taurus/core/util/event.py
@@ -172,14 +172,19 @@ class EventGenerator(Object):
         else:
             pass
 
-    def subscribeEvent(self, cb, data=None):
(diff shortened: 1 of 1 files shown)


reszelaz:
Sardana will need it also for suppor-3.x. I have checked and a cherry pick applies without problems.

cpascual:
> Sardana will need it also for support-3.x.

Could you create another PR for that?

reszelaz:
Done. See #483.

PR #486, opened 2017-07-04
Apply same as #426 in support-3.x

cpascual:
Fix exception when connecting to a Tango 9.2 DB.
This is a backport to taurus 3 of the #426 PR


reszelaz:
I'm accepting this PR because the change must be safe but I could not test it - I don't have a non-mysql Tango database.

PR #487, opened 2017-07-04
Use full name when checking if a DB is mysql

cpascual:
The cache refresh does a check to see if a mysql optimization can be done.
This is done by checking the tango DB device, which is obtained using a call 
to `TangoAuthority.dev_name`. The` dev_name` method is exposed from the 
Tango API and returns a tango URI **without the authority component**. 
In the current implementation this may lead to checking the wrong device in some 
special cases. Avoid it by prefixing the tango URI with the authority full name.


reszelaz:
Looks safe to me. Merging it. Thanks!

PR #489, opened 2017-07-06
Add configuration properties to TaurusTrend2D (Fixes #250) 

### LINKED ISSUE #250: TaurusTrend2D: config properties missing
(from an internal request at ALBA)

It would be nice if TaurusTrend2DDialog widget would store config info on, at least, the following settings:

The following settings should be handled as taurusconfigproperties so that they can be recovered when reopening a GUI
* autoscale setting for all axes ( x,y and z)
* scale min and max for all axes
* interpolation and colormap settings (parameters/image menu)
* axes settings (title, font, ...)

cpascual:
Refers to #250 

Make TaurusTrend2D store/retrieve the following config properties:

- Axes properties: limits, scale, title , color, unit
- lut_range
- interpolation
- colormap
- Stack mode
- Model
- x, y and z autoscale status (it now uses the new autoscale tools)


cpascual:
For facilitating the test, you can use the following snippet:

```python
from taurus.qt.qtgui.extra_guiqwt import TaurusTrend2DDialog
from taurus.core.taurushelper import changeDefaultPollingPeriod

if __name__ == '__main__':
    changeDefaultPollingPeriod(300)
    import sys
    app = TaurusApplication()

    w = TaurusTrend2DDialog(autoscale='xy')

    model = 'eval:rand(30)'
    w.setModel(model)

    try:
        w.loadConfigFile('/tmp/kk.cfg')
        pass
    except Exception as e:
        print "Exception loading config:", e

    w.show()

    ret = app.exec_ ()

    w.saveConfigFile ('/tmp/foo.cfg')

    sys.exit(ret)
```

cpascual:
@taurus-org/integrators : can someone review & integrate this?

teresanunez:
@taurus-org/integrators, I will have a look now.

teresanunez:
@cpascual, your example doesn't work in my computer. Could be the guiqwt version?.
 I get the error:

   w = TaurusTrend2DDialog(autoscale='xy')
  File "/usr/lib/python2.7/dist-packages/taurus/qt/qtgui/extra_guiqwt/taurustrend2d.py", line 84, in __init__
    self.add_tool(toolklass)
  File "/usr/lib/python2.7/dist-packages/guiqwt/plot.py", line 277, in add_tool
    tool.register_plot(plot)
  File "/usr/lib/python2.7/dist-packages/taurus/qt/qtgui/extra_guiqwt/tools.py", line 298, in register_plot
    baseplot.SIG_ITEMS_CHANGED.connect(self.items_changed)
AttributeError: 'ImagePlot' object has no attribute 'SIG_ITEMS_CHANGED'

reszelaz:
I have the same problem. I use guiqwt 3.0.3.

teresanunez:
My version of guiqwt is 2.3.1

cpascual:
@ teresa,: You are right, my changes only work on for guiqwt>=3 (what it does not make sense to me is that @reszelaz is getting the same).

I'll have a look.

BTW: since this is a change for Taurus4, I propose to raise the requirement of guiqwt to >=3 because guiqwt 2.3.1 uses old-style signals. and also we can get rid of ugly try-excepts in the code. What do you think?

reszelaz:
If it helps, I use Debian9 official package:
```
In [1]: import guiqwt

In [2]: guiqwt.__version__
Out[2]: '3.0.3'

In [3]: guiqwt.__file__
Out[3]: '/usr/lib/python2.7/dist-packages/guiqwt/__init__.pyc'

In [4]: from guiqwt.signals import SIG_ITEMS_CHANGED
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-4-60b67fb5e553> in <module>()
----> 1 from guiqwt.signals import SIG_ITEMS_CHANGED

ImportError: cannot import name SIG_ITEMS_CHANGED

```

reszelaz:
Ups, I was using a wrong branch. There are no problems with guiqwt 3. Sorry...

teresanunez:
@cpascual, if a version >= 3.0  is needed for some of the new features, yes, I think it is better to add
the requirement.

cpascual:
I just increased the dependency on guiqwt (for which I also updated from develop).
@teresanunez : can you try again with guiqwt 3?

teresanunez:
@cpascual , yes, I will try it today or tomorrow.

teresanunez:
Hi @cpascual, sorry, I can not try it with guiqwt3 ... our system administrators try to install it in one
 computer, but it breaks too many despendencies.

cpascual:
Hi @teresanunez , what do you propose?, should someone else ( @reszelaz) integrate it?, or should we reconsider supporting guiqwt 2.3.1? (If so, I would prefer treating the support of guiqwt<2.3.1 as a request to be done *after* July17...)

teresanunez:
Hi @cpascual, thanks for asking. I would propose that @reszelaz integrate this. About the support
 for guiqwt 2.3.1 I would not try to implement it now and only would do it if it is really needed at 
some point.

reszelaz:
The functional test revealed one problem: the interpolation is not correctly restored from the configuration. The rest of the features works well. Whenever this is fixed I will proceed with the merge cause I don't have any comments to the code.

All the features I've tested with a dedicated taurusgui and the configuration stored in the ini file. Just the stack mode was tested by the snippet that you have posted.

While testing taurustrend2 I have realized about two issues:
- some of the actions e.g. "Parameters" are disabled (context menu, toolbar) until you either move the image or resize the window.
- changing the curve titles also requires resizing the window to apply the changes.

cpascual:
> the interpolation is not correctly restored from the configuration. The rest of the features works well. Whenever this is fixed I will proceed with the merge cause I don't have any comments to the code.

I'll look into it now.

> some of the actions e.g. "Parameters" are disabled (context menu, toolbar) **until you either move the image or resize the window.**

I think you are not totally right in the last part of the sentence... those actions are disabled unless you select an item for which they make sense (in this case the image). When you select it they are enabled. This is standard guiqwt behaviour, not a bug (although I agree that it is annoying when you are unaware of it)

> changing the curve titles also requires resizing the window to apply the changes.

I do not understand what you refer to. What do you refer to with "curve titles"?

reszelaz:
> those actions are disabled unless you select an item for which they make sense (in this case the image)

It could be, I remember that I had to somehow manipulate the widget/application in order to get them active. Thanks for the explanation!

> What do you refer to with "curve titles"?

Sorry, I wanted to say "axes titles".

cpascual:
I just fixed the issue with the interpolation.

Regarding the axes title refresh issue, I just checked and it seems to be a bug in guiqwt's axes tool
You can reproduce it with: `python -m guiqwt.tests.image`. I already reported it in guiqwt: https://github.com/PierreRaybaut/guiqwt/issues/76

PR #491, opened 2017-07-10
Add qtpy mock when building docs in RTD (Fixes #490)

### LINKED ISSUE #490: (doc) RTD fails to generate taurus.qt.qtgui API docs
Hello, 
Link broken in the web site. See the issue in the next link:

http://www.taurus-scada.org/en/latest/devel/api/taurus/qt/qtgui.html

Access to the API doc's in 'latest ' doc version is broken.

cpascual:
RTD fails to build the taurus.qt.qtgui API because it cannot import
qtpy. Add a dummy qtpy module to sys.modules when we detect that we are
under RTD to fix this (i.e., implement workaround "b)" from https://github.com/taurus-org/taurus/issues/490#issuecomment-314079369)

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -42,9 +42,27 @@ sys.path.append(_mock_path)
 
 # insert mock for qtpy (only when building docs in RTD)
 # see https://github.com/taurus-org/taurus/issues/490
+# ... and since we are at it, add also a few other modules to workaround
+# more RTD failures to build API
+# (recipe inspired in https://stackoverflow.com/a/35229746 )
 if os.environ.get('READTHEDOCS') == 'True':
-    import types
-    sys.modules['qtpy'] = types.ModuleType('qtpy', 'Mock of qtpy for RTD')
+    from mock import MagicMock
+    MOCK_MODULES = ['qtpy',
+                    'qtpy.QtWidgets',
+                    'epics',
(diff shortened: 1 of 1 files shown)


reszelaz:
I've checked this solution locally and it worked for me, so I assume that in the RTD environment it will also work. 

Just for the curiosity, why you insert `types.ModuleType` instead of `mock.MagicMock`?

cpascual:
>Just for the curiosity, why you insert types.ModuleType instead of mock.MagicMock?

Both would have worked. I chose types because it is in the stdlib (although we know that RTD has mocks installed) and because in this case qtpy is not used at all, just needs to be imported. In fact I could have duplicated any other module (e.g. `sys.modules['qtpy']=sys` would have worked also !)

cpascual:
Note: I tested this using an empty virtualenv where I declared the READTHEDOCS envvar and run sphinx:

```
$ mkvirtualenv rtd
(rtd) $ export READTHEDOCS='True'
(rtd) $ pip install sphinx
(rtd) $ pip install sphinx_rtd_theme
(rtd) $ pip install mock
(rtd) $ sphinx-build -T -E -b html -d _build/doctrees-readthedocs -D language=en doc/source/ _build/html
```

reszelaz:
Works for me as well. Merging...

PR #496, opened 2017-07-12
TaurusValue bckcomp of label configs

reszelaz:
This PR aims to fix the backwards compatibility of the TaurusValue label configs.
The old-style label configs are now correctly translated to the new-style fragments e.g. <attr_name> -> name,  <dev_fullname>: parentObj.fullname.

**Note**: `TaurusValue.setLabelConfig` used with a simple string was previously setting a label text to this string. IMHO this was working just by coincidence. For me the `setLabelConfig` method should not be used to achieve this kind of behavior. That's why this PR does not maintain the backwards compatibility here.

### DIFF
--- a/lib/taurus/qt/qtgui/display/tauruslabel.py
+++ b/lib/taurus/qt/qtgui/display/tauruslabel.py
@@ -231,6 +231,7 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
     def __init__(self, parent=None, designMode=False):
         self._prefix = self.DefaultPrefix
         self._suffix = self.DefaultSuffix
+        self._permanentText = None
         self._bgRole = self.DefaultBgRole
         self._fgRole = self.DefaultFgRole
         self._modelIndex = self.DefaultModelIndex
@@ -398,6 +399,10 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
     def resetSuffixText(self):
         self.setSuffixText(self.DefaultSuffix)
 
+    def setPermanentText(self, text):
+        self.setText(text)
+        self._permanentText = text
+
     def setAutoTrim(self, trim):
         self._autoTrim = trim
         self.controllerUpdate()
@@ -422,6 +427,11 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
 
     def resetAutoTrim(self):
         self.setAutoTrim(self.DefaultAutoTrim)
+
+    def displayValue(self, v):
+        if self._permanentText is not None:
+            return self._permanentText
+        return TaurusBaseWidget.displayValue(self, v)
 
     @classmethod
     def getQtDesignerPluginInfo(cls):
--- a/lib/taurus/qt/qtgui/display/tauruslabel.py
+++ b/lib/taurus/qt/qtgui/display/tauruslabel.py
@@ -252,6 +252,11 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
         if self._designMode:
             self.controllerUpdate()
 
+        # register configurable properties
+        self.registerConfigProperty(
+            self.getPermanentText, self.setPermanentText, "permanentText"
+            )
+
     def _calculate_controller_class(self):
         ctrl_map = _CONTROLLER_MAP
         if self._designMode:
@@ -398,6 +403,9 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
 
     def resetSuffixText(self):
         self.setSuffixText(self.DefaultSuffix)
+
+    def getPermanentText(self):
+        return self._permanentText
 
     def setPermanentText(self, text):
         self.setText(text)
--- a/lib/taurus/qt/qtgui/panel/taurusvalue.py
+++ b/lib/taurus/qt/qtgui/panel/taurusvalue.py
@@ -1216,6 +1216,20 @@ class TaurusValue(Qt.QWidget, TaurusBaseWidget):
 
     @Qt.pyqtSlot('QString')
     def setLabelConfig(self, config):
+        """Sets fragment configuration to the label widget.
+
+        :param config: fragment
+        :type config: str
+        """
+        # backwards compatibility: this method used to work for setting
+        # an arbitrary text to the label widget
+        try:
+            self.getModelFragmentObj(config)
+        except Exception:
+            msg = "Use setLabelText for setting an arbitrary label text"
+            self.deprecated(msg)
+            self.setLabelText(config)
+            return
         self._labelConfig = config
         self.updateLabelWidget()
 
--- a/lib/taurus/qt/qtgui/panel/taurusvalue.py
+++ b/lib/taurus/qt/qtgui/panel/taurusvalue.py
@@ -106,11 +106,11 @@ class DefaultLabelWidget(TaurusLabel):
             TaurusLabel.setModel(self, None)
             self.setText(devName)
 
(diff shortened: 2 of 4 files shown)


lib/taurus/qt/qtgui/panel/taurusvalue.py
```diff
@@ -106,11 +106,11 @@ def setModel(self, model):
             TaurusLabel.setModel(self, None)
             self.setText(devName)
 
-    _BCK_COMPAT_TAGS = {'<attr_name>': '{attr.name}',
-                        '<attr_fullname>': '{attr.fullname}',
-                        '<dev_alias>': '{dev.name}',
-                        '<dev_name>': '{dev_name}',
-                        '<dev_fullname>': '{dev.fullname}',
+    _BCK_COMPAT_TAGS = {'<attr_name>': 'name',
```
cpascual: I think that `TaurusValue.resetLabelConfig` needs also be changed for coherence with this

lib/taurus/qt/qtgui/panel/test/test_taurusvalue.py
```diff
@@ -51,9 +51,8 @@ def test_bug126(self):
         '''Verify that case is not lost when customizing a label (bug#126)'''
         w = self._widget
         # self._widget.setModel('eval:1')
-        self._widget.setModel('tango:' + DEV_NAME + '/double_scalar')
+        self._widget.setModel('tango:' + DEV_NAME + '/MIXEDcase')
```
cpascual: I may be wrong, but I think that the changes that you are proposing in this test are not ok.

The original purpose of this test was to check for bug sdn-SF#126 (corresponding to #53 ): we want to make sure that the case is kept when *setting* the label.

With your proposed changes, we are checking that a mixed label is not lost, which IMHO is a good thing to check, but in a different test.

If you agree, can you please implement both tests? (and in the meanwhile, it would be nice to change the name and docstring to use `#53` instead of `#126`)

cpascual:
Hi @reszelaz , will you have time to address my comment above?

reszelaz:
Ups, I completly forgotten about it. I will do it this week.

lib/taurus/qt/qtgui/panel/taurusvalue.py
reszelaz: Done

lib/taurus/qt/qtgui/panel/test/test_taurusvalue.py
reszelaz: I agree. I had to add a new features to the wiedgets:
* TaurusLabel - setPermanentText
* TaurusValue - setLabelText

I have also implemented backwards compatibility to the setLabelConfig so the original test_bug126 works.

lib/taurus/qt/qtgui/panel/taurusvalue.py
```diff
@@ -1212,13 +1216,31 @@ def getLabelConfig(self):
...
+        """Sets fragment configuration to the label widget.
+
+        :param config: fragment
+        :type config: str
+        """
+        # backwards compatibility: this method used to work for setting
+        # an arbitrary text to the label widget
+        try:
+            self.getModelFragmentObj(config)
+        except Exception:
```
cpascual: Is it not possible to narrow this exception to AttributeError?

lib/taurus/qt/qtgui/panel/test/test_taurusvalue.py
```diff
@@ -76,6 +77,20 @@ def texts(self, model=None, expected=None, fgRole=None, maxdepr=0):
         self.assertEqual(got, expected, msg)
         self.assertMaxDeprecations(maxdepr)
 
+    def test_labelCaseSensitivity(self):
+        '''Verify that case is respected of in the label widget'''
```
cpascual: PEP8: Better use double-quotes for docstrings (even if in the other methods we are wrongly using single-quotes)

lib/taurus/qt/qtgui/panel/test/test_taurusvalue.py
```diff
@@ -76,6 +77,20 @@ def texts(self, model=None, expected=None, fgRole=None, maxdepr=0):
...
+        '''Verify that case is not lost when customizing a label (bug#126)'''
```
cpascual: Is this line a leftover from a copy-paste?

lib/taurus/qt/qtgui/panel/taurusvalue.py
reszelaz: What we get there is a simple Exception:

```
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/panel/test/test_taurusvalue.py", line 56, in test_bug126
    w.setLabelConfig(label)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/panel/taurusvalue.py", line 1227, in setLabelConfig
    self.getModelFragmentObj(config)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/base/taurusbase.py", line 631, in getModelFragmentObj
    return self.modelObj.getFragmentObj(fragmentName)
  File "/home/zreszela/workspace/taurus/lib/taurus/core/taurusmodel.py", line 201, in getFragmentObj
    obj = getattr(obj, fn)
  File "/home/zreszela/workspace/taurus/lib/taurus/core/tango/tangoattribute.py", line 303, in __getattr__
    % name)
Exception: TangoAttribute does not have the attribute MIXEDcase

```

lib/taurus/qt/qtgui/panel/test/test_taurusvalue.py
cpascual: Maybe a PEP8 commit fixing all the quotes in this file would be great

lib/taurus/qt/qtgui/panel/test/test_taurusvalue.py
```diff
@@ -76,6 +77,20 @@ def texts(self, model=None, expected=None, fgRole=None, maxdepr=0):
...
+        w = self._widget
+        # self._widget.setModel('eval:1')
```
cpascual: This line seems to be a leftover from debugging... remove it?

lib/taurus/qt/qtgui/display/tauruslabel.py
```diff
@@ -231,6 +231,7 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
     def __init__(self, parent=None, designMode=False):
         self._prefix = self.DefaultPrefix
         self._suffix = self.DefaultSuffix
+        self._permanentText = None
```
cpascual: This new property should be registered as a configproperty

lib/taurus/qt/qtgui/panel/taurusvalue.py
```diff
@@ -1212,13 +1216,31 @@ def getLabelConfig(self):
...
+            return
         self._labelConfig = config
         self.updateLabelWidget()
 
     def resetLabelConfig(self):
-        self._labelConfig = '{attr.label}'
+        self._labelConfig = 'label'
         self.updateLabelWidget()
 
+    def setLabelText(self, text):
```
cpascual: This should be registered as a configproperty

lib/taurus/qt/qtgui/display/tauruslabel.py
reszelaz: Done

reszelaz:
I think that now I have addressed all the code review comments.

cpascual:
Yes, merging. Thanks!

PR #499, opened 2017-07-13
Fix #484: jdrawview fails to draw unless resized

### LINKED ISSUE #484: TaurusJDrawSynopticView sometimes does not start drawing
This is similar to #360 (already fixed), but not quite the same...

To reproduce, download the [syntest.txt](https://github.com/taurus-org/taurus/files/1109437/syntest.txt) file and run:
`python -m taurus.qt.qtgui.graphic.jdraw.jdraw_view syntest.txt`

Most of the times (although not always) I get an empty white synoptic:

fail

If I resize the window, a value appears and from that moment it correctly updates every 3s (without requiring further interaction):

afterresizing

Sometimes the value is shown from the beginning and it updates ok, so it may be a race condition with events or signals...

cpascual:
A workaround implemented in `TaurusJDrawSynopticsView.fitting` causes
the #484 bug. Fix it by avoiding to call `fitting` with
ADJUST_FRAME=True in `setModel`. Also add TODO to consider deprecating
the workaround code altogether, since it seems to be causing more
problems than it solves.

Fixes #484

### DIFF
--- a/lib/taurus/qt/qtgui/graphic/jdraw/jdraw_view.py
+++ b/lib/taurus/qt/qtgui/graphic/jdraw/jdraw_view.py
@@ -214,10 +214,17 @@ class TaurusJDrawSynopticsView(Qt.QGraphicsView, TaurusBaseWidget):
         self.debug('\toffset = %s,%s ; size = %s,%s' % (x0, y0, w, h))
         self.fitInView(x0, y0, w, h, Qt.Qt.KeepAspectRatio)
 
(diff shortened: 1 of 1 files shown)


cpascual:
Te actual fix was proposed internally by @cmft and I reviewed it, so I am integrating it myself

sergirubio:
image

Tested taurus develop with Vacca on elinac demo; it works

cmft:
Ok. Thanks @sergirubio !

PR #508, opened 2017-07-13
Store geometry in config of TaurusPlot & TaurusTrend

### LINKED ISSUE #502: --config option for taurusform and taurusplot does not keep window size
Hi again,
it seems --config option does not save window size so I need to resize window at every launch. 
I suspect this is the behaviour of all similar taurus command line options.

Is it a normal behaviour? In this case, what is the correct method to save window size?

cpascual:
TaurusPlot widget does not sav/restore the window geometry. Add
geometry to the config properties of TaurusPlot and restore it
if it is a top level window

Fixes #502


cpascual:
To test it:

Run taurusplot, change window size and/or position, save the settings as `foo.pck`
Then open another taurusplot using `taurusplot --config foo.pck`

The same also works for taurustrend

cpascual:
@taurus-org/integrators could you intetgrate it ASAP so that it gets in Jul17 release ?

PhilLAL:
Thanks you !

LJBD:
@cpascual I've tested it and it restores the size of the window. I'm merging this.

PR #509, opened 2017-07-14
Use displayValue to format the datainspector label

### LINKED ISSUE #181: Taurustrend DataInspector data format
Hi,

We would need the DataInspector tool from taurustrend either:
- to use the data format of the Tango DB or 
- to have the possibility to configure it from taurustrend application

The problem appears when trying to plot the RF Frequency of the Accelerators, which have 9 significant numbers (499.654.125 Hz) and it is configured at Tango DB as %16.4f. When we plot it with taurustrend it is not possible for us to know the frequency value using data inspector because it only shows the first 5 most significant numbers (4.9965e08).

This feature is necessary for us, specially when we have to retrieve the frequency value after som machine failures.

Thanks in advance,
Marc

Reported by: marcalvarez ( http://sf.net/u/marcalvarez )

cpascual:
DataInspector mode in TaurusPlot  uses a hardcoded format for 
displaying the Y value.
This causes issues when more precisionis required (see
https://github.com/taurus-org/taurus/issues/181 )
Fix it by delegating the formatting to `displayValue()` (this requires
to have access to the associated trendset from the curve when 
working on a TaurusTrend)

Fixes #181 

Known limitations: 
- The inspector does not react to changes in format while it is running (but new instances of the plot/trend will use the correct format)
- The x value is not treated like the Y. This is a consequence of the X model being a "second-class citizen" (there is no TaurusBaseComponent associated to it and corespondingly no `displayValue()` method)

### DIFF
--- a/lib/taurus/qt/qtgui/plot/taurusplot.py
+++ b/lib/taurus/qt/qtgui/plot/taurusplot.py
@@ -2975,6 +2975,8 @@ class TaurusPlot(Qwt5.QwtPlot, TaurusBaseWidget):
                             pickedCurveName = name
                             pickedIndex = i
                             pickedAxes = curve.xAxis(), curve.yAxis()
+                            _displayValue = getattr(curve, 'owner', curve
+                                                    ).displayValue
         finally:
             self.curves_lock.release()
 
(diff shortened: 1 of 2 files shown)


reszelaz:
I have tested it with taurustrend and taurusplot and in both cases it works nicely. I was not able to see the first limitation that you list above. In my case the inspector reacts on the display format changes in the attribute configuration without the need to restart anything. Merging it...

PR #514, opened 2017-07-19
Add deprecation warnings

mrosanes:
Add deprecation warnings for tauruscurve and taurustrend1d.

This widgets are deprecated, instead of them, taurustrend should
be used. Add deprecation warnings.


cpascual:
I agree with the deprecation of these two, but could you please use self.deprecated (and use its API) instead of self.info?

Also, the please add tauruscurve and taurustrend1d to the deprecated section of this release in the changelog

mrosanes:
Usage of self.deprecated applied.
Updated changelog accordingly.

PR #525, opened 2017-07-20
fixing typo and menu changes in users doc

PhilLAL:
That's just minor typo correction, and also some menu titles which seem to have changed.
I did not find a contrib for doc so I hope I've done this with respect to the standard you follow.

I plan to update this PR as I find new typos or menu titles changes.

Issue #527, opened 2017-07-20
[taurusgui] how to include a JDraw synoptic


PhilLAL:
Hello,
following the [doc](taurus-scada.org/en/latest/users/ui/taurusgui.html#creating-custom-panels), I understood it is possible to include a panel which is a JDraw application already developped. 
I tried to include this JDraw with "Add Panel" selecting Widget Type=TaurusJDrawSynopticsView but I do not have anywhere to add my JDraw location.

However, it seems to work when adding this synoptic using taurusgui --new-gui and selecting JDW synoptic file.
How can we add a JDraw synoptic into an existing TaurusGui?

cpascual:
There is no current way of doing it while the application is launched, but you can use the "new gui wizard" to edit an existing TaurusGui:

For example:
- create a new gui with `taurusgui --new-gui` and use the `/tmp/foo` dir as its project dir (first input of the wizard). Let the wizard finish and check that the gui exists and works.
- Now, to edit it, call again `taurusgui --new-gui` and select again `/tmp/foo` dir as the project dir. You will be asked if you want to edit the existing GUI. Say yes and you can modify the GUI (including adding/removing synoptics). Note that no changes will be saved unless you accept in the last page of the wizard.

PS: while testing this for this reply, I discovered that #477 introduced a regression that prevents to edit existing GUIs (a fix has been proposed in #532)

PhilLAL:
Thank you for this explanation.
In understand the "Add panel" into the Gui does not allow to add a Widget Type=TaurusJDrawSynopticsView

Then, perhaps would it be more clear to comment the code making this widget appearing into the "Add panel" window? It could be always uncommented when adding a Synoptic will be possible after the creation. Isn'it ?

cpascual:
> perhaps would it be more clear to comment the code making this widget appearing into the "Add panel" window? 
You are right.

> It could be uncommented when adding a Synoptic will be possible after the creation. Isn'it ?

Actually, it should be easy to modify the synoptics widget to accept the file name using a "file:" scheme name in its setModel method and this would allow the file to e passed with the Advanced settings in the "add panel" dialog.

PhilLAL:
Thank you, it works!
Thus this procedure should be documented since it seems not straightforward that Model can be used to enter a JDW file (I understood from http://taurus-scada.org/en/latest/users/ui/modelchooser.html it is used for device/attributes).

cpascual:
> Thus this procedure should be documented 

[...]

PR #539, opened 2017-07-21
(doc) Document how to add synoptic panel

PhilLAL:
solve #527


cpascual:
Hi @PhilLAL thanks for the PR,

I see that you created this *on top of* the same branch that you used for #525. Therefore, by merging this PR we would also be merging the changes of 51d3e75 and d039e25, apart from the one properly related to this PR ( 0d7873f )
If that is intentional, no problem: we can accept this PR and close #525. But if it is not intentional, please close this one and resubmit from a branch where only the changes from 0d7873f are done.

PhilLAL:
No, in fact, I should have divided this into two PR, one for just 0d7873f, and another one for the doc typos.
I will make the modifications to make this clean.

PR #540, opened 2017-07-21
Details to add a synoptic panel after taurusgui creation.

PhilLAL:
Add into doc some explanations about how to add a synoptic panel.
I made the choice to add this into taurusgui.rst and not modelchooser.rst

solve #527


cpascual:
Merged after a minor change. Thanks!

Issue #544, opened 2017-07-27
[doc] users/ui misplaced paragraph


PhilLAL:
Hello, 
into doc/source/users/ui/index.rst (at the end of the page) appears the following sentences:
"This section explains some features that are common to most applications built with taurus. This is done from the GUI user point of view (not for developers).
For a detailed list of features of each widget, please refer to the Developer’s Guide"

However, it seems to me this sentences would be more understandable if it appears at the beginning of the users/ui chapter to explain what it contains. 
Perhaps is it due to the fact that toctree sphinx instruction add file contents where each part is listed.

cpascual:
Closed via #546

PR #546, opened 2017-07-27
Doc users ui intro

PhilLAL:
Solve #544


PhilLAL:
Since this PR is based on a new branch created after my commit on doc (bad idea), it should be applied only after  #525 is applied.

cpascual:
Merged. 
Thanks for contributing (especially on the docs which tend to get less attention)!

Issue #545, opened 2017-07-27
[doc] misplaced figures in the pdf


PhilLAL:
Hi,
I noticed that some figures  of the PDF doc are misplaced outside from their chapter. For example,taurus configuration browser figure (http://taurus-scada.org/en/latest/users/ui/configurations.html) and  screenshots figures (http://taurus-scada.org/en/latest/users/screenshots.html) are placed into developer guide section into the PDF doc.

Perhaps some anchors to figures could help (I think to something like https://stackoverflow.com/questions/13706250/linking-to-figures-in-sphinx or https://stackoverflow.com/questions/2686310/referencing-figures-with-numbers-in-sphinx-and-restructuredtext)
Also, some new pages (https://stackoverflow.com/questions/26120570/sphinx-add-a-page-break-in-my-pdf-with-latexpdf) could help also. However in this case, it will probably move taurus config browser figure to screenshots section (except if a newpage is set at the end of every section, but I do not think it is not a good choice).

cpascual:
Thanks for reporting!
If you can test some of the proposed solutions and make a PR, we would really appreciate it!

PhilLAL:
I would need to get into the doc building to check it works, but the only try I have done was not conclusive.
I can suggest a PR without build if someone can check...

cpascual:
> I can suggest a PR but without building it, it will need someone who can check... Does it sound possible?

Definitely. That would be great!

(Only one thing: I will be out of office next 2 weeks, so I won't be able to look into it until I come back)

PhilLAL:
Ok (I will be back at the same time, note sure I will have time to do this before).

To help (me) doing tests, here are some informations:

- concerned instructions: ".. figure" or ".. image" (not sure why there is two different instructions used)
- figures which are misplaced into the pdf:

  * figure /_static/taurusmodelchooser01.png appears into the toc of "TaurusPlot User’s Interface" section instead of "TaurusModelChooser User’s Interface" section
  * figure /_static/taurusconfigbrowser.png appears into "Developer's Guide" section instead of "Taurus Configuration Browser User’s Interface" section
  * figure /_static/taurusdemo.png appears into "Taurus Configuration Browser User’s Interface" section instead of "Taurus Configuration Browser User’s Interface" section
  * figures /_static/taurusgui-newpanel01.png and /_static/taurusgui-permanentpanels01.png appear into "Perspectives" subsection instead of "Creating custom panels" subsection
[...]

PR #547, opened 2017-07-28
Misplaced figures

PhilLAL:
try to test #545


PhilLAL:
Precision: the doc has to be built to check figures are well placed in the PDF since I have not been able to build it.

cpascual:
Please merge the latest develop into this branch (solving whatever conflicts) so that we can test it

cpascual:
@PhilLAL : I pushed this to be built in the sandbox.
The result will eventually be available in: https://readthedocs.org/projects/taurus-sandbox/downloads/pdf/phillal-misplaced-figures/

cpascual:
This does not seem to do the trick. 

The solution seems to be to use captions in the figures (so that it does not matter if the figure gets placed too late in the pdf (any help would be appreciated)

See https://github.com/taurus-org/taurus/issues/545#issuecomment-323327916 .

So I am closing this PR and leaving #545 open until someone takes the task of writing captions...

Issue #507, opened 2017-07-13
(doc) PDF table of contents nearly empty


PhilLAL:
The PDF toc is nearly empty, it just contains : 

>1    Projects related to Taurus    3
>1.1 Taurus 4.0 documentation    3
>Python Module Index               579

cpascual:
I confirm it. But I have no idea on how to fix it (this is sphinx-magic done at readthedocs). Help would be appreciated.
As a workaround, if you want to have a "printer-friendly" doc you could use the Epub file (the TOCs are well generated for the epub and it is also not affected by #498 )

PhilLAL:
I checked sphinx but I am not familiar enough to help for this, sorry.
Perhaps the ":hidden:" just after the ".. toctree::" is linked to this issue...

cpascual:
Fixed by #548

PR #548, opened 2017-07-28
Doc hidden toctree

PhilLAL:
Try to solve #507 
It seems :hidden: option under the toctree directive causes table of content to be hidden.


PhilLAL:
Precision: the doc has to be built to check table of content is visible since I have not been able to test it.

cpascual:
>  the doc has to be built to check if table of content is visible in the PDF since I have not been able to test it.

Hi, I pushed these changes to abranch called rtdtests in my personal fork:

https://github.com/cpascual/taurus/tree/rtdtests

This branch is configured as the source of a "sandbox" RTD project that I sometimes use to test RTD-related changes before applying them to the official repo. You can check the generated docs in:

http://taurus-sandbox.readthedocs.io/en/rtdtests/

(just be aware that I reuse this branch for different things, so do not base any work on it)

cpascual:
Note: at the moment of writing this, the build has finished (see http://readthedocs.org/projects/taurus-sandbox/builds/5841340/ ) but for some reason the pdf and epub files are not visible in the Downloads section of RTD. 
I'll allow a bit more time just in case it takes some time migrating the artifacts...

cpascual:
Something is not properly configured in the sandbox... the pdf should be available in 
http://readthedocs.org/projects/taurus-sandbox/downloads/ but it is not. 
I'll see what I can do...

PhilLAL:
I read page "i" of generated document https://media.readthedocs.org/pdf/taurus-sandbox/latest/taurus-sandbox.pdf but it did not changes anything.

cpascual:
Now it worked. The page 1 of  https://readthedocs.org/projects/taurus-sandbox/downloads/pdf/rtdtests/ now shows a depth 3 TOC:

image

(and the same TOC is browsable in the PDF viewer)

PhilLAL:
Geat, it seems however the number of items showed is not the good one, isn't it?
I'll try incrementing by one the level of toctree.

cpascual:
So we know that this can be done, but now we need to decide what is the best approach.

First some considerations: 

- At the moment we are using sphinx to generate both the "Taurus Main Page" (index.html) and the documentation proper (docs.html).
- I **personally** prefer the index.html to be a small (ideally single-screen) page. Tha means no TOC on the main page. 
- For the TOC to be useful in the PDF it should have at least 3 depth levels **from `docs.rst`** (currently, this PR sets it at 3 levels from `index.rst`).

It seems to me that the best option would be if we could somehow force the PDF and epub documentation to be generated with `docs.rst` as its root  (instead of `index.rst`).

If that is not possible, we could consider reorganizing the docs a bit to merge `docs.rst` and  `index.rst`, but we should be careful in order to not break external source's links (e.g. from issues, emails, bookmarks,...) to the docs. In other words: a link such as http://www.taurus-scada.org/en/latest/devel/api/taurus/core.html should be valid after whatever changes we do.

What do you think?
Any other opinions from @taurus-org/integrators or anybody else?

PhilLAL:
Indeed, docs maxdepth is set to 2 but could be set to 3.
However, I do not know how sphinx manage an inconsistent maxdepth level between index.rst and included docs.rst.

cpascual:
Hi, I just checked using *both* `:hidden:` and a maxdepth of 4 and we get exactly what we want  (see https://readthedocs.org/projects/taurus-sandbox/downloads/pdf/rtdtests/ )

... except for the detail that the toc should be at the top of the index, not below to avoid having everything under "projects related to Taurus"

So I am going to check it and if so I'll propose the final changes in this same PR

cpascual:
I just pushed some changes to index.rst (248ca9e) . IMHO, the resulting PDF has a reasonable TOC (see the image below). Is this ok for you, @PhilLAL ?

image

PhilLAL:
Sounds great to me ! :)

Issue #498, opened 2017-07-13
UI colors not printed in the pdf


PhilLAL:
Hi there,
I am not sure this is desired effect but in the source/users/ui/ui_colors.rst, the ".. raw:: html" command before each color tab make it printed only in html doc.
For an offline reading of the doc, a similar view would be useful.

edit : colors can be found int the html doc at page http://taurus-scada.org/en/latest/users/ui/ui_colors.html

cmft:
Hi @PhilLAL ,
You are right. Could you create a PR?

Many thanks for reporting it.

PhilLAL:
I am not sure how to organize same data in a PDF but I'll see when I get some time.

PhilLAL:
I checked sphinx but I am not familiar enough to help for this, sorry.

cpascual:
Just as we did for #549 & #585, I am closing this since it relates to PDF generation and we don't do that anymore

PR #549, opened 2017-07-28
Doc ui colors

PhilLAL:
Just a try to test colors into PDF to solve #498


PhilLAL:
Just a precision: doc need to be build to check the colors of "Taurus Device state colors" section are visible in the PDF doc.

cpascual:
> Just a precision: doc need to be build to check the colors of "Taurus Device state colors" section are visible in the PDF doc.

I built the docs with this changes. You can find it in : 

http://readthedocs.org/projects/taurus-sandbox/downloads/pdf/phillal-doc-ui-colors/

But I think that the proposed changes do not solve the issue...

cpascual:
@PhilLAL , can you confirm that this is not working?

If so, do you have any more ideas, or should I close the PR?

PhilLAL:
I confirm, it does not work. I will look for some possible ideas during this week...

PhilLAL:
I confirm, it does not work. I will look for some possible ideas during this week...

PhilLAL:
I asked to some collegues likely to help but but did not get any solution. I do not see how to do this, a temporary solution could be to print a table without the color, at least it will print the name of the colors.

cpascual:
While I certainly I love the idea of getting rid of the latex in the docs, this precise workaround sacrifices colors in html docs for consistency with PDF. My impression is that the html is used a lot more than the pdf, and therefore it would not be a good workaround. Also, if accepted, the "Preview" column would make no sense.

But it looks like these objections could be eliminated using something like [this](https://stackoverflow.com/a/9510628) 
Or even something based on [this](https://stackoverflow.com/a/9753677) (which apparently works for pdf as well):

PhilLAL:
Indeed, preview lose all interest in PDF.
Another solution could be to use the RST table only for PDF document. It will let user know the colors but without printing the color in the table.

Also, I was trying to get the exact appearance of the HTML doc for the PDF, but using only font color instead of background color in the cell would be another solution. It seems the solution proposed in the 2d link you proposed if I'm not wrong.

cpascual:
Hi @PhilLAL , this has been stalled for some time. And given that we are not producing the PDF anymore, it is kind of obsolete.
So I am closing the PR for now as rejected (but feel free to reopen at any time if you feel motivated to try some other implementation)

PhilLAL:
No problem, thanks for informing me @cpascual .

PR #554, opened 2017-08-23
Use new docker tags in Travis.yml

cpascual:
The docker tags in taurus-test docker image have been updated. Now
`debian-{jessie,stretch,buster}` are used insted, while "debian9" and
"latest" should no longer be used. Modify travis.yml accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,21 @@
 language: python
 
+python:
+  - "2.7"
+
 sudo: required
 
+services:
+  - docker
+
+env:
+  - DOCKER_IMG=cpascual/taurus-test:debian-jessie
+  - DOCKER_IMG=cpascual/taurus-test:debian-stretch
+  - DOCKER_IMG=cpascual/taurus-test:debian-buster
+
 matrix:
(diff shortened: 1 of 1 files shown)


cpascual:
The travis tests are [failing for jessie](https://travis-ci.org/taurus-org/taurus/jobs/267606793) due to some problem with spyder / qtconsole.

Note that ipython-qtconsole has been *removed* during the [jessie image build](https://hub.docker.com/r/cpascual/taurus-test/builds/bi68dhkqyrkbn7jbxj4aptf/)

cpascual:
I am stuck here. I cannot find the way of satisfying spyder >=3 in debian8.

So, unless someone proposes a better action, I propose to relax the dependency on spyder by providing a dummy TaurusBaseEditor in case of trouble. In fact the code is already present (but commented) in `taurus/qt/qtgui/editor/__init__.py`

cpascual:
I am merging this one since the changes proposed work well. The travis failures are related with the debian-jessie image, and should be solved elsewhere.

Issue #497, opened 2017-07-12
[Travis] aborted builds (10m without output)


cpascual:
We are lately seeing many spurious failures in travis where a test gets stuck (maybe an event lost?).
These are gone when retrying.

I open this issue to keep a record of these (the history is lost when restarting the build).
I propose to log them here in comments in order to find patterns.

cpascual:
See: https://travis-ci.org/taurus-org/taurus/builds

At least the following builds have failed (although they may now be ok due to manually triggered rebuids):
270.2 , 273.2 276.2, 282.2, 284.2

Note that it is always the test on the debian9 docker (the ".2")

The test in which it fails varies, but it seems to be in taurus.core.tango.test.test_tangoattribute.AttributeTestCase  (although this may just be a coincidence due to timing)

A typical log in case of error ends as:
```
test_write_read_conf (taurus.core.tango.test.test_tangoattribute.AttributeTestCase)

Testing write_read_conf(expected=[<Quantity(100, 'dimensionless')>, <Quantity(300, 'dimensionless')>], cfg='alarms', attr_name='short_scalar_nu', value=[100, 300]) ... 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

cpascual:
I wonder if it may be that Travis only considers "stdout" for output and maybe the test suite is using stderr... (or something like this). To check it we could try to redirect the output when calling testsuite...

reszelaz:
I have observed something very similar with the sardanatestsuite. At some point, after numerous subscription & unsubscription to the tango attributes, the events stops arriving to the tests. This is 100% reproducible with the sardanatestsuite on Debian9 with the standard distribution packages. I have investigated it a little bit. For example, the events are correctly emitted by the server - another client running in parallel to the tests receives them. But finally I had to stop the investigation due to the lack of time....

In sardana we don't suffer it yet at Travis cause the sardanatestsuite docker image is still based on the Debian8.

cpascual:
It seems to me that, after a period when this happened a lot, it is not happening anymore (or at least it happens much less often).

Maybe it is related with the changes in Travis infrastructure related to #601 ...

[...]

PR #556, opened 2017-08-24
Relax spyder dependency

cpascual:
Installing spyder>=3 can be a problem in older systems (e.g. debian 8)
because it forces many dependencies (e.g. ipython>4) which may not be
wanted. On the other hand, spyder is only needed for taurus.qt.editor.

Allow importing taurus.qt.editor even if spyder is not installed.
In that case, return a dummy widget and log a warning. 
Also make the extras_requires target for spyder more specific.

This will fix the errors seen in the travis builds since #554 was merged.

### DIFF
--- a/lib/taurus/qt/qtgui/editor/__init__.py
+++ b/lib/taurus/qt/qtgui/editor/__init__.py
@@ -30,11 +30,7 @@ __docformat__ = 'restructuredtext'
 try:
    from .tauruseditor import *
 except Exception as e:
-   from taurus.qt.qtgui.display import create_fallback as __create
-   TaurusBaseEditor = __create("TaurusBaseEditor")
    from taurus import warning, debug
-   warning('Problem with taurus.qt.editor.\n'
-           ' A dummy TaurusBaseEditor will be used\n' +
-           ' (maybe you do not have spyder >=3 ?). ')
+   warning('Problem with taurus.qt.editor (hint: is spyder >=3 installed?)')
    debug('%r', e)
 
--- a/lib/taurus/qt/qtgui/editor/__init__.py
+++ b/lib/taurus/qt/qtgui/editor/__init__.py
@@ -27,10 +27,14 @@
 
 __docformat__ = 'restructuredtext'
 
-from .tauruseditor import *
+try:
+   from .tauruseditor import *
+except Exception as e:
+   from taurus.qt.qtgui.display import create_fallback as __create
+   TaurusBaseEditor = __create("TaurusBaseEditor")
+   from taurus import warning, debug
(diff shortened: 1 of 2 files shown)


cpascual:
The tests are passing now (*) for all distros (jessie, stretch, buster).
@taurus-org/integrators : so could some @taurus-org/integrators check if this solution is acceptable and if so integrate it? 

(*) Note: actually, it took **many** (>5) rebuilds to get the [Travis build for stretch](https://travis-ci.org/taurus-org/taurus/jobs/267900804) . For some reason it was either affected by #497 or some similar spurious failures (sometimes related to CORBA timeouts, but on other occasions I have the impression that the problems were directly caused by a Travis bug ...)

cmft:
I try this but it does not work when spyder 3 is not installed.

```
import sys
from taurus.qt.qtgui.editor import TaurusBaseEditor

def demo():
        w = TaurusBaseEditor()
        w.resize(900, 800)
        w.show()
        w.load(__file__)
        return w

if __name__ == "__main__":
        app = TaurusApplication()
        w = demo()
        sys.exit(app.exec_())
```

cpascual:
> I try this but it does not work when spyder 3 is not installed.
> ...
> w.load(__file__)

Yes, but that is expected. The dummy is quite dummy ;)
The purpose is just to avoid exceptions when importing the module or initializing the TaurusBaseEditor class or using it as a placeholder in TaurusDesigner. Other than that I expect it to fail.

Of course it could be improved (e.g. by subclassing from MagickMock) but I feel that it is not wrth going into such lengths.

cmft:
> Yes, but that is expected. The dummy is quite dummy ;)

You're right. I did a test without loading, and it works. 

But during the test I saw that the dummy widget does not add nothing, 
so I believe that warning message is enough and avoid to create it (I will remove these lines)

``` 
+   from taurus.qt.qtgui.display import create_fallback as __create
+   TaurusBaseEditor = __create("TaurusBaseEditor")
```
After these changes the PR can be integrated.

cpascual:
> But during the test I saw that the dummy widget does not add anything,
> so I believe that warning message is enough and avoid to create it (I would remove these lines)
> 
> +   from taurus.qt.qtgui.display import create_fallback as __create
> +   TaurusBaseEditor = __create("TaurusBaseEditor")
> 

Removing the dummy TaurusBaseEditor is also a good solution for me. It would still allow the editor submodule to be imported and the warning to be issued... and the result is probably cleaner.

So I am doing a commit removing the dummy.

cpascual:
If no one disagrees, I will merge this (with @cmft's blessings)

cpascual:
2 days passed,no one complained... merging.

PR #558, opened 2017-08-24
Make taurusargparse tango independent

mrosanes:
Make taurusargparse tango independent.

Modify --tango-host by --authority option.
This gives the option to enter authorities other than Tango, even
though, as of today, only a Tango authority will have influence
using this option.

### DIFF
--- a/lib/taurus/core/util/argparse/taurusargparse.py
+++ b/lib/taurus/core/util/argparse/taurusargparse.py
@@ -53,7 +53,7 @@ Currently, the known options are:
 
     #. ``--help`` prints the total number of available options
     #. ``--taurus-log-level`` sets the taurus log level
-    #. ``--tango-host`` sets the default tango host
+    #. ``--authority`` sets the default tango host
     #. ``--taurus-polling-period`` sets the default taurus global polling period (milliseconds)
     #. ``--taurus-serialization-mode`` sets the default taurus serialization mode
     #. ``--remote-console-port`` enables remote debugging using the given port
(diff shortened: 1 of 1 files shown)


cpascual:
I do not thing this is the right thing to do because:

- Currently there is no need for the generic "--authority" option (we do not foresee using it in any other scheme)
- The resulting code is more complex (and fragile), and it is still tango-centric (the only difference is cosmetic, because the tango-centricity is hidden to the user)

IMHO the option should be left as it is, and eventually (when [TEP13](http://www.taurus-scada.org/en/latest//tep/?TEP13.md) is done), we could consider adding an entry point so that schemes can insert their own argparse options (and then move this code to the taurus-tango plugin)

mrosanes:
I think that it is not only cosmetic, the new code could be easily extended to be used for other schemes other than Tango if it was required in the future, by using the same option --authority. Even if today only the Tango option has sense. As you say the mechanism will be better once [TEP13](http://www.taurus-scada.org/en/latest//tep/?TEP13.md) is implemented, but I think the contribution is removing the Tango centricity, because it opens the door to introduce the other schemes in the same block of code if required ('if/elif/else').

cpascual:
> the contribution is removing the Tango-centricity, because it opens the door to introduce the other schemes in the same block of code if required ('if/elif/else').

I do not agree. The Tango-centricity issue is still there: `TaurusApplication` (which should consist only of scheme-agnostic code) is still doing tango-specific code. The only difference is that in your proposed code, the user does not see it with an explicit tango-centric arg name. 

Furthermore, the contributed solution invites developers to make things worse by adding *other-scheme-specific* code where currently there is "only" tango-specific code. This is not only undesirable, but it also has clear scalability limitations (keep in mind that there may be third-party schemes). 

And regarding the "opening the door" part, I think it is not right. The door was already open since one could just as well add extra argparse options (e.g. `--eval-auth`) if really required (and BTW, this makes more sense because, in that situation, one could want to set different default authorities for different schemes)

cpascual:
Nobody added to this discussion in more than 1 month, so I followed my criterion and rejected it. Of course, if someone considers that it should be re-considered, please reopen it.

PR #559, opened 2017-08-29
Fix Taurus may emit events with outdated data

### LINKED ISSUE #216: Taurus may emit events with outdated data (Tango scheme)
In the Tango scheme there are three occassions that that may update the attribute ( **\__attr_value**, **\__attr_err**, **\__attr_timestamp**). These methods are:

* **read** (executed in the main thread)
* **push_event** (executed in the Tango event consumer thread)
* **poll** (executed in the Taurus polling thread)

The last two of them emit Taurus events.

Current implementation may lead to emitting events with the outdated values.  The polling mechanism reads the Tango attributes asynchronously (first drop the read requests and later on collect the replies). The more up to date event value may get overriden by an older value corresponding to the polling request dropped prior to the event reception.

We have observed that in a particular GUI with a lot of attributes being polled. One of the polled attributes was receiving events at the same time. The TaurusLabel with that attribute was changing to the value corresponding to the event, but suddently returing to show the previous value. After approx. 3 seconds the label refreshed correctly.

I propose to treat the manipulation of attribute members ( \__attr_value, \__attr_err, \__attr_timestamp) as a critical section and protect access to it with a lock. Furthermore the update should be discarded if the value present in the attibute is more up to date.

Reported by: reszelaz ( http://sf.net/u/zreszela )

### LINKED ISSUE #226: Avoid deprecation warning from IPython 4.x
Avoid using features deprecated by IPython 4.x
(e.g. IPython.config)

cmft:
Current implementation may lead to emitting events with
the outdated values. The polling mechanism reads the Tango
attributes asynchronously (first drop the read requests
and later on collect the replies).

The more up to date event value may get overriden by an
older value corresponding to the polling request dropped
prior to the event reception.

Implement a filter at TangoAttribute level to drop older events.

This filter can be activated/deactivated via a variable in
TaurusCustomSettings

Fix #216


lib/taurus/core/tango/tangoattribute.py
```diff
@@ -448,8 +449,17 @@ def poll(self, **kwargs):
             if single:
                 self.read(cache=False)
             else:
-                self.__attr_value = self.decode(kwargs.get('value'))
+                value = self.decode(kwargs.get('value'))
                 self.__attr_err = kwargs.get('error')
+                filter_old_event = getattr(tauruscustomsettings,
+                                           'FILTER_OLD_TANGO_EVENTS', False)
+                if self.__attr_value is not None and \
```
cpascual: PEP8 note: better use parenthesis instead of `\`

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -448,8 +449,17 @@ def poll(self, **kwargs):
...
+                        self.__attr_err is None and \
+                        filter_old_event and \
+                        kwargs.get('time') < self.__attr_value.time.totime():
```
cpascual: I see a potential problem in this implementation (i do not know if in practice it is important):
`kwargs.get('time') < self.__attr_value.time.totime()` will return True if kwargs does not have the `time` key.  
I think that in this case one should **not** discard the event. Maybe it should give an exception (using `kwargs['time']`), or it should return False (using `kwargs.get('time', float('inf'))`)

lib/taurus/core/tango/tangoattribute.py
cpascual: Also, this filter block should be properly documented, since its justification is not obvious

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -727,8 +737,17 @@ def _pushAttrEvent(self, event):
                  evt_value is a TaurusValue, an Exception, or None.
         """
         if not event.err:
-            self.__attr_value, self.__attr_err = self.decode(
-                event.attr_value), None
+            attr_value = self.decode(event.attr_value)
+            filter_old_event = getattr(tauruscustomsettings,
+                                       'FILTER_OLD_TANGO_EVENTS', False)
+            if self.__attr_value is not None and \
```
cpascual: PEP9 use parenthesis instead of `\`

cpascual:
Also, the changes should be documented in the changelog

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -436,21 +436,26 @@ def write(self, value, with_read=True):
...
-                        self.__attr_err is None and \
-                        filter_old_event and \
-                        kwargs.get('time') < self.__attr_value.time.totime():
+
+                # Discard "valid" notifications (value is not None and error
+                # is None) if FILTER_OLD_TANGO_EVENTS is enabled
+                # and the given timestamp is older than the timestamp
+                # of the cache value
+                if (self.__attr_value is not None
+                    and self.__attr_err is None
```
cpascual: The check of self.__attr_err is None could be avoided if we put this filter after the block of line 517

cmft:
You can find a  test  for the critical section [here](https://gist.github.com/cmft/cd4cc517db45b5264aed73dac1c929c2)

Issue #408, opened 2017-02-23
[doc] Get rid of mocks for docs


cpascual:
The docs generation requires importing all modules of taurus for introspection.

This forced us to generate mocks for the various taurus dependencies... but this is an ugly hack which is not easy to maintain (see, e.g. #267).

I see two different options:

a) Move the document creation to Travis.

b) Make use of the now-available [conda support in RTD](http://docs.readthedocs.io/en/latest/conda.html) to install the dependencies using Conda 

I prefer a) because I think that it is easier to maintain in the longer term (we maintain the docker images for the CI anyway), but this needs a bit more initial work since we need to configure the artifact deployment in travis to upload the docs somewhere else.

On the other hand, b) may be less disruptive and quicker to implement on the short term. But I personally have no experience with Conda... so any help would be welcome (@bilderbuchi?  , @pgap ?)

bilderbuchi:
If you have to mock purely because to make the import taurus pass so that sphinx can grab the docstrings, couldn't you get away with a MagicMock(), and a trick like [this](http://stackoverflow.com/a/8658332/599884) to actually mock away the imports of your dependencies? I.e. something like (untested)
```
import sys
sys.modules['pytango'] = MagicMock()  # and the others
import taurus
```
Regarding b) are all the dependencies of taurus and its dependencies (looking at pytango and Tango, here) available on conda? I have just tried to install Taurus on Windows on an Anaconda environment, and it didn't work out, I needed at least Tango and the VC++ Compiler for Python 2 from the general internet, probably boost, too, in the end.

regardig a) uploading from travis to RTD seems to be an open issue: rtfd/readthedocs#1083

cpascual:
> couldn't you get away with a MagicMock()

Actually our mocks are using MagicMock, but the  needed implementation (in order to meet the documentation requirements) is a *lot* more complex than the simple examples. And, what is worse, for many modules, it requires special tweaks. If you are curious, have a look at the buildmock.py script in the doc/ dir

> are all the dependencies of taurus and its dependencies (looking at pytango and Tango, here) available on conda?

I think they are (but I've never used conda myself). I think that PyTango uses conda for CI (maybe @tiagocoutinho or @vxgmichel can confirm). Also, if I remember correctly, someone ( was it  @pgap? ) told me that pyepics was also conda-installable. 

> uploading from travis to RTD seems to be an open issue rtfd/readthedocs.org#1083

[...]

PR #562, opened 2017-09-01
Issue408: Add gh-pages support

mrosanes:
Add gh-pages support to be able to use
the dockers taurus-test for building documentation, 
and be able to get rid of mocks for building the docs 
later on.

Edit: This is a proposal to deal with #408 . The results can be seen in https://mrosanes.github.io/taurus/


.travis.yml
```diff
@@ -20,7 +20,19 @@ matrix:
 before_install:
   - docker run -d --name=taurus-test -h taurus-test --volume=`pwd`:/taurus $DOCKER_IMG
   - sleep 10
+  - docker exec taurus-test /bin/bash -c "apt-get -qq update"
+  - docker exec taurus-test /bin/bash -c "apt-get install -y python-sphinx-rtd-theme"
```
cpascual: I would do this (install sphinx-rtd-image) in the docker images instead.

.travis.yml
```diff
@@ -20,7 +20,19 @@ matrix:
...
   
 script:
   - docker exec taurus-test /bin/bash -c "cd taurus ; python setup.py install"
-  - docker exec taurus-test /bin/bash -c "TAURUS_STARTER_WAIT=5 taurustestsuite -e 'taurus\.core\.util\.test\.test_timer'"
+  - travis_wait docker exec taurus-test /bin/bash -c "TAURUS_STARTER_WAIT=5 taurustestsuite -e 'taurus\.core\.util\.test\.test_timer'"
+  - docker exec taurus-test /bin/bash -c "cd taurus ; python setup.py build_sphinx"
+  - docker exec taurus-test /bin/bash -c "cd taurus/build/sphinx/html ; touch .nojekyll"
```
cpascual: Maybe it would be better to avoid the `cd`: 
```
docker exec taurus-test /bin/bash -c "touch /taurus/build/sphinx/html/.nojekyll"
```

cpascual:
Make sure that the deploy only happens with one image (I suggest debian-stretch).
This can be done [using `condition`](https://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A):



Also, currently the

cpascual:
Note that the resulting docs lack the graphs created with graphviz (it may be related to a missing dependency in the images). Compare, e.g.: https://mrosanes.github.io/taurus/devel/api/taurus/qt/qtgui/display/_TaurusLabel.html#tauruslabel with http://www.taurus-scada.org/en/latest/devel/api/taurus/qt/qtgui/display/_TaurusLabel.html#tauruslabel

.travis.yml
mrosanes: I agree, but I don't know if I have access to modify the taurus-test docker files. @cpascual, could you add the python-sphinx-rtd-theme to the dockers?

cpascual: I just added it to debian-stretch and debian-buster images.
I am not sure we should add it to debian-jessie

cpascual:
This approach was good as a proof of concept, but it has several issues:

- It uses a personal token (which would grant write travis access to *all* repos to which the user has access) instead of a deploy key
- It deploys to the same repo (again, a security problem)
- It does not support version-control of the docs nor publishing more than one version of the docs

These issues have been addressed in #572. Therefore I am closing this PR and we can continue the work on #572 and #408

Issue #563, opened 2017-09-01
Allow to set formatter in the taurus GUIs 


cmft:
Nowadays Taurus allows to set custom formatters programmatically but the default widgets (taurusform, taurustrend, ...), use the default Formatter and does not allow to change it.

This enhancement propose:
- Add menu action to choice/set a Formatter
- Allow to configure the default Formatter  (e.g. via tauruscustomsetting)

MikeFalowski:
Hey! What is status of this? I was just looking for engineering notification in taurusform.

cpascual:
Oh, we forgot to close this ticket when we solved it in #564.

Basically now (in the develop branch / coming release) you can change the formatter in the following ways:

1. (GUI) set a custom formatter in TaurusForm (globally or per attribute) via a context menu option. TaurusForms will store this preference in their configs.
2. (GUI/PROG) use the `TaurusBaseWidget.onSetFormatter()` slot to set a custom formatter on any other widget
3. (PROG) set a custom formatter on any taurus widget (using `TaurusBaseComponent.setFormat`)
4. (PROG) set a custom formatter on any taurus class (by overwriting `.FORMAT`)
5. (CLI) set the defaultFormatter via a TaurusApplication command line argument (`--default-formatter`)
6. (CONF) set the defaultFormatter via tauruscustomsettings.DEFAULT_FORMATTER


Note that options 5 and 6 deal with the **default** formatter and therefore will affect all widgets (except those that explicitly changed it) and therefore should be used with great care and only with very generic formatters. In other words: Don not use those options unless you **really** know what you are doing. Option 4 should also be used with care if applied to a base class (since all child classes will be affected)

PR #564, opened 2017-09-05
Allow to set custom Formatter

cmft:
Add API to set custom formatter. 
Add an action to set a formatter in some taurus widget
This enhancement propose:
 * Add menu action to set a Formatter (taurusvalue, taurusform and taurusplot/trend)
 * Allow to configure the default Formatter via tauruscustomsetting

#563


lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -1243,6 +1243,48 @@ def __init__(self, name, parent=None, designMode=False):
...
+        finally:
+            return formatter
+
+    def showFormatterDlg(self):
+        '''
+        showFormatterDlg show a dialog to get the formatter from the user.
+        :return: formatter: python fromat string or formatter callable
+        (in string version) or None
+        '''
+        formatter, ok = Qt.QInputDialog.getText(self, "Set formatter",
```
cpascual: The dialog should show the current formatter.
And ideally the e dialog could be implemented with an editable combobox  (a Qt.QInputDialog.getItem) instead of a lineedit and pre-populate it with supported formatters.

lib/taurus/tauruscustomsettings.py
```diff
@@ -68,6 +68,14 @@
...
 # EXTRA_SCHEME_MODULES = ['myownschememodule']
 
+# Set a custom formatter. Taurus widgets use a default formatter based on the
+# attribute type, but sometimes a custom formatter is needed.
+# The formatter can be a python format string or formatter callable
+# (in string version)
+#e.g.
+# DEFAULT_FORMATTER = '{0}'
+# DEFAULT_FORMATTER = 'taurus.core.tango.util.formatter.tangoFormatter'
+
```
cpascual: Consider adding a `--default-formatter` option in the argument parser of TaurusApplication which would override `tauruscustomsettings.DEFAULT_FORMATTER`

lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -1243,6 +1243,48 @@ def __init__(self, name, parent=None, designMode=False):
...
+            moduleName, formatterName = formatter.rsplit('.', 1)
+            __import__(moduleName)
+            module = sys.modules[moduleName]
+            formatter = getattr(module, formatterName)
+        except:
+            pass
+        finally:
+            return formatter
+
+    def showFormatterDlg(self):
```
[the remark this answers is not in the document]
cmft: It is not merged  cause some widgets need a special implementation of  `onSetFormatter` meanwhile the `showFormatterDlg` is the same implementation for all of them.

lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -1243,6 +1243,48 @@ def __init__(self, name, parent=None, designMode=False):
         self.call__init__(TaurusBaseComponent, name,
                           parent=parent, designMode=designMode)
         self._setText = self._findSetTextMethod()
+        formatter = getattr(taurus.tauruscustomsettings,
```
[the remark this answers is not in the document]
cmft: I had  the same dilemma when I implement it. I did it because `TaurusBaseComponent` has a programmatic API to set the "formater" (`setFormat()`) via python module or via python format (string). 
While the `formater`  only accept string (Something that only should come as a view input according my implementation purpose).

I did it to not mix both API after checking  that "all" (official) taurus widget inherit from `TaurusBaseWidget`. But there is not major reason to not move it

lib/taurus/qt/qtgui/base/taurusbase.py
cmft: I  believe that the first requirement can be easily done.
But the use of combobox (for the moment) is very complicated since the formaters are not registered (as plugin ) and we allow to use custom external modules.

lib/taurus/tauruscustomsettings.py
cmft: Agree with the `--default-formatter` option for individual widgets (form, plots, ...) 
but, what does it mean override for you? More precedence? or replace it for all the widgets? 

I think it is difficult to  manage in complex GUI e.g.  taurusgui where it should affect to all widgets

But IMHO it is a extra since it can be done by a menu action.

lib/taurus/qt/qtgui/base/taurusbase.py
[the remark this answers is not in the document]
cpascual: fair enough.

lib/taurus/qt/qtgui/base/taurusbase.py
cpascual: I agree. let's do the first part and leave a TODO for the second

lib/taurus/tauruscustomsettings.py
cpascual: Implementationwise I was thinking that --default-formatter would simply overwrite at runtime the`tauruscustomsettings.DEFAULT_FORMATTER` value

Then any widget in the app which does not explicitly set a formatter, would use this one. If an app needs finer control, they should neither be using this parameter nor the setting.

lib/taurus/qt/qtgui/base/taurusbase.py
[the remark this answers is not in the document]
cpascual: The rule of thumb for me is: if an API makes sense when running with a QObject (e.g. by a QGraphicsObject) , then it should be in TaurusBaseComponent. If the API specifically uses QWidget calls, then it goes to TaurusBaseWidget.
I think that in your implementation, some of the methods belong to one case and some to the other

PR #572, opened 2017-09-14
Make travis build docs and publish them

cpascual:
This is related to #408 . It makes travis build the docs and deploy them to taurus-org/taurus-doc repo.

This implementation (which uses [doctr](https://drdoctr.github.io)) is inspired on the work by @mrosanes in #562 but it addresses the following issues that were missing there:

- Use a deploy_key instead of a personal token (the deploy key is safer since it won't give write access to *all repos* in the taurus org)
- Use of a separate repo for docs (again, this is more secure since travis is only granted write access to the doc serving repo, not to the one holding the code)
- Keep the changes under version control
- Build docs in separate dirs for each branch or tag of the taurus repo (this allows us to serve docs for older versions and also to easilly test doc changes). The develop branch is built on the root, so we

The results can be seen, in:

 `https://taurus-org.github.io/taurus-doc/` 

... for the develop version (*only after we merge this PR*) and in 

`https://taurus-org.github.io/taurus-doc/v-BRANCH_OR_TAG_NAME/` 

... for other tags or branches in the taurus repo (for example, see  https://taurus-org.github.io/taurus-doc/v-doctr )

**Important**: note that in this PR I am *not* disabling the parts needed by RTD, so the docs are still built by RTD in parallel and served in their usual location, while the docs built by travis get served by GH pages. This would allow us merge the PR without breaking RTD and test and compare before deciding to point our custom domain to GH pages instead of RTD.


cpascual:
@taurus-org/integrators Can anyone have a look and integrate this?

teresanunez:
Hi @cpascual, I have had a look and I agree with the ideas and how the documentation
is created in:  
https://taurus-org.github.io/taurus-doc/v-doctr
I don't know if you want us to check something more before integrating, if not I could make it.

cpascual:
There is only one thing that I expect that could potentially be wrong: I am not sure if the `.nojekyll` file will be kept in the root of the docs when we merge this PR. But that can only be tested by actually merging. If it is not kept, the gh pages docs won't be visible, but it will be just a matter making sure that the file is set (I would make a direct push afterwards).

In conclusion, please go ahead with the merge if you did not find any other problem.

cpascual:
>  I am not sure if the .nojekyll file will be kept in the root of the docs when we merge this PR

Issue #576, opened 2017-09-21
ExternalApp(...) creates zombie processes of exit


jkotan:
Hi,

our users complain that `ExternalApp(...)` , e.g. `xterm spock` or `pymca`, when you open it and close it many time produces a lot of zombie processes ( i.e. `<defunct>`). If the  user GUI is running for longer time it can be even 100  or more zombies which is annoying.
 A possible fix would be from time to time to remove zombies, e.g.

    diff --git a/lib/taurus/qt/qtgui/util/taurusaction.py b/lib/taurus/qt/qtgui/util/taurusaction.py
    index 1c46911..b01d836 100644
    --- a/lib/taurus/qt/qtgui/util/taurusaction.py
    +++ b/lib/taurus/qt/qtgui/util/taurusaction.py
    @@ -109,6 +109,7 @@ class ExternalAppAction(Qt.QAction, BaseConfigurableClass):
		 if any(args):
		     #Qt.QMessageBox.warning(self.parentWidget(),'Warning','In ExternalAppAction(%s)'%args)
		     self._process.append(subprocess.Popen(args))
    +                self._process = [p for p in self._process if p.poll() is None]
		     return True
		 else:
		     return False

What do you think?

Best regards,
Jan

cpascual:
Hi Jan,
The proposed solution seems reasonable (but I haven't tested it)
Would you care to create a PR?

cpascual:
Fixed by #577

PR #577, opened 2017-09-21
remove zombies on start new ExternalApp

jkotan:
Hi,

It is a fix for  issue #576.

Bests,
Jan


jkotan:
Weird. For me the errors does not seem to be related to the fix.

cpascual:
> the errors does not seem to be related to the fix
 
you are right. One was related to #497  and the other was more of the #261 type...

It is a shame that we haven't ye being able to get rid of these. I just forced a relaunch

cpascual:
Sorry, I forgot come back to merge after confirming that the tests were ok. Merging now

PR #580, opened 2017-09-26
Taurus reports false warnings

### LINKED ISSUE #579: Taurus reports false warnings
Attributes that are not quantities should not report unitless warning.

e.g. 

```
taurusform  sys/tg_test/1/state

MainThread     WARNING  2017-09-26 09:00:57,296 controls01.cells.es:10000.sys/tg_test/1.state: Unknown unit "No unit (will be treated as unitless)"

taurusform  sys/tg_test/1/boolean_scalar

MainThread     WARNING  2017-09-26 09:04:05,820 controls01.cells.es:10000.sys/tg_test/1.boolean_scalar: Unknown unit "No unit (will be treated as unitless)"

```

cmft:
Attributes that are not quantities report unitless warning.

Fix #579


cpascual:
The implementation seems to be triggering some testsuite errors: https://travis-ci.org/taurus-org/taurus/jobs/279836470

cmft:
The  problems are with the `uchar_scalar` attributes; Taurus considered these attributes numeric and PyTango not. So, the implementation of `isNumeric` of the tango scheme is incoherent. We  should  take a  desicion about  what is  wrong, the tests or the `isNumeric` implementation.

cpascual:
After conversation with @tiagocoutinho  (PyTango), we decided that we should consider uchar_scalar attributes as **non-numeric** in Taurus (since this will be more aligned with them being considered "bytes" in python 3.

Several unit tests may need to be modifiedaccordingly, and possibly the behaviour of TaurusValue, TaurusValuesTable and TaurusImageDlg should be checked for consistency.

cpascual:
The final implementation from @cmft  is somewhat a middle point: the uchar tango attrs are considered `bytes` (therefore, not numerical), but the I/O is done viewing them as `uint8` (just as PyTango does).

I think that this is fine. Merging

PR #587, opened 2017-10-07
Enable unit test for write/read epics attribute

cpascual:
The epics.AttributeTestCase.write_read_attr was disabled because
epics was not working in the CI test docker images.
Reactivate it because epics seems to work well now.


cpascual:
Note, that 85a7bd3 may be (partially) solving  #497

PR #589, opened 2017-10-11
Fix strange behavior of Tango validator

### LINKED ISSUE #586: Strange behavior of Tango validator with strict=False
I observe a strange behavior of the new Tango validators, well actually just tested on the device validator, but it could affect others as well.

With taurus/support-3.x the validator behaves as following:

```
In [1]: from taurus.core.taurusvalidator import DeviceNameValidator

In [2]: v = DeviceNameValidator()

In [3]: v.getNames("controls01:10000/a/b/c")
Out[3]: ('a/b/c', 'a/b/c', '')
```

With taurus/develop the validator behaves as following:

```
In [1]: from taurus.core.tango.tangovalidator import TangoDeviceNameValidator

In [2]: v = TangoDeviceNameValidator()

In [3]: v.getNames("controls01:10000/a/b/c")

In [4]: v.getUriGroups("controls01:10000/a/b/c")

In [5]: v.getUriGroups("controls01:10000/a/b/c", strict=False)
```

So in all these cases, the validator returns nothing. Shouldn't in in case of strict flag set to False return something?

cmft:
Some back. comp. URIs do not work. The Tango validator does not
return the group when the strict mode is set to False.
    
Fix #586

### DIFF
--- a/lib/taurus/core/tango/tangovalidator.py
+++ b/lib/taurus/core/tango/tangovalidator.py
@@ -182,7 +182,8 @@ class TangoDeviceNameValidator(TaurusDeviceNameValidator):
                   r'(\?(?P<query>%(query)s))?' + \
                   r'(#%(fragment)s)?$'
         authority = '(?P<host>([\w\-_]+\.)*[\w\-_]+):(?P<port>\d{1,5})'
-        path = '/?(?P<devname>((?P<_devslashname>[^/?#:]+/[^/?#:]+/[^/?#:]+)))'
+        path = '/?(?P<devname>((?P<_devalias>(?<=//)([^/?#:]+))|' + \
+               '(?P<_devslashname>[^/?#:]+/[^/?#:]+/[^/?#:]+)))'
 
         return pattern % dict(scheme=self.scheme,
                               authority=authority,
--- a/lib/taurus/core/tango/tangovalidator.py
+++ b/lib/taurus/core/tango/tangovalidator.py
@@ -176,7 +176,19 @@ class TangoDeviceNameValidator(TaurusDeviceNameValidator):
         '''In non-strict mode, allow double-slash even if there is no Authority.
         (e.g., "tango://a/b/c" passes this non-strict form)
         '''
-        return self.namePattern.replace('tango):(', 'tango)://(')
+        pattern = r'^((?P<scheme>%(scheme)s)://)?' + \
+                  r'((?P<authority>%(authority)s)(?=/))?' + \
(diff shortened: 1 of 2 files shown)


cpascual:
The proposed changes are triggering unit test failures: 

```
FAIL: test_isValid_117 (taurus.core.test.test_taurushelper.IsValidNameTestCase)

Testing isValid(strict=False, elementType=3, name='tango://alias')

  File "/usr/local/lib/python2.7/dist-packages/taurus-4.1.2_alpha-py2.7.egg/taurus/test/base.py", line 113, in newTest

  File "/usr/local/lib/python2.7/dist-packages/taurus-4.1.2_alpha-py2.7.egg/taurus/core/test/test_taurushelper.py", line 222, in isValid

    self.assertEqual(returned, expected, msg)

AssertionError: isValidName returns False (expected True for name=tango://alias and etype=Device)

```

lib/taurus/core/tango/tangovalidator.py
```diff
@@ -250,14 +263,16 @@ def nonStrictNamePattern(self):
         """
 
         # allow for *optional* double-slashes and *optional* ?configuration...
```
cpascual: This comment does not apply anymore

lib/taurus/core/tango/tangovalidator.py
```diff
@@ -176,7 +176,20 @@ def nonStrictNamePattern(self):
...
         (e.g., "tango://a/b/c" passes this non-strict form)
         '''
-        return self.namePattern.replace('tango):(', 'tango)://(')
+        pattern = r'^((?P<scheme>%(scheme)s)://)?' + \
+                  r'((?P<authority>%(authority)s)(?=/))?' + \
+                  r'(?P<path>%(path)s)' + \
+                  r'(\?(?P<query>%(query)s))?' + \
+                  r'(#%(fragment)s)?$'
+        authority = '(?P<host>([\w\-_]+\.)*[\w\-_]+):(?P<port>\d{1,5})'
+        path = '/?(?P<devname>((?P<_devalias>(?<=//)([^/?#:]+))|' + \
```
cpascual: The lookbehind for the double-slash here is reproducing a bug in taurus3, not a feature. It is true that the latest implementation in Taurus3 would not allow `tango://alias`, but that is a "bug" (since according to the taurus3 docs it should be correct). In these cases we decided that the backwards compatibility layer should follow (if possible) the doc rather than the implementation.

cpascual:
Looks ok to me. Merging

PR #593, opened 2017-10-24
Add TangoFactory option to disable event subscribing

sergirubio:
This is needed to bypass the Tango 9 slowness when connecting to hundreds of attributes. It doesn't affect default Taurus behaviour, just provides a flag that can be set by clients to disable the event subscribing if needed.

The changes are:
 - added set_tango_events_disabled() method to TangoFactory
 - used by TangoAttribute to subcribe or not to Tango Events
 - explicit call to DeviceProxy.subscribe_event moved to a separate method

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -618,7 +618,7 @@ class TangoAttribute(TaurusAttribute):
             self._activatePolling()
             self._call_dev_hw_subscribe_event(True)
                 
-    def _call_dev_hw_subscribe_event(self,stateless=True):      
+    def _call_dev_hw_subscribe_event(self, stateless=True):      
         try:
             attr_name = self.getSimpleName()
             cid = self.__dev_hw_obj.subscribe_event(
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -287,7 +287,9 @@ class TangoAttribute(TaurusAttribute):
 
         # subscribe to configuration events (unsubscription done at cleanup)
         self.__cfg_evt_id = None
-        self._subscribeConfEvents()
+        self._ignore_events = self.factory().get_tango_events_disabled()
+        if not self._ignore_events:
+            self._subscribeConfEvents()
 
     def cleanUp(self):
         self.trace("[TangoAttribute] cleanUp")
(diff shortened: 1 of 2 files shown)


lib/taurus/core/tango/tangoattribute.py
```diff
@@ -603,24 +605,36 @@ def _subscribeEvents(self):
...
+            self._activatePolling()
+            return
+
         try:
             self.__subscription_state = SubscriptionState.Subscribing
-            self.__chg_evt_id = self.__dev_hw_obj.subscribe_event(
-                attr_name, PyTango.EventType.CHANGE_EVENT,
-                self, [])  # connects to self.push_event callback
-
+            self._call_dev_hw_subscribe_event(False)
```
cpascual: I am not sure about this change. Note that the implementation of `_call_dev_hw_subscribe_event` handles all exceptions, whereas the original code may raise exceptions on errors, which would in turn trigger the code in the except clause of this try-except block.
It seems to me that the new implementation makes that the except clause is never reached.

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -603,24 +605,36 @@ def _subscribeEvents(self):
...
+        except:
+            self.error(traceback.format_exc())
+                
     def _unsubscribeEvents(self):
         # Careful in this method: This is intended to be executed in the cleanUp
         # so we should not access external objects from the factory, like the
         # parent object
-        if self.__dev_hw_obj is not None and self.__chg_evt_id is not None:
+        
+        if self.__dev_hw_obj is not None and self.__chg_evt_id not in (None,-1):
```
cpascual: Can you explain why the check of `-1` here?

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -603,24 +605,36 @@ def _subscribeEvents(self):
...
+        except:
+            self.error(traceback.format_exc())
+                
     def _unsubscribeEvents(self):
         # Careful in this method: This is intended to be executed in the cleanUp
         # so we should not access external objects from the factory, like the
         # parent object
-        if self.__dev_hw_obj is not None and self.__chg_evt_id is not None:
+        
+        if self.__dev_hw_obj is not None and self.__chg_evt_id not in (None, -1):
```
cpascual: Can you explain why checking `-1` here?

lib/taurus/core/tango/tangofactory.py
```diff
@@ -182,6 +183,16 @@ def get_default_tango_host(self):
         """Retruns the current default tango host
         """
         return self._default_tango_host
+    
+    def set_tango_events_disabled(self,value):
```
cpascual: IMHO, it would be better if the property was named tango_events_enabled:

set_tango_events_enabled
is_tango_events_enabled

sergirubio:
amended

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -600,26 +609,33 @@ def _subscribeEvents(self):
                 self.debug("failed to subscribe to chg events: HW is None")
                 return
 
-        self.__subscription_event = threading.Event()
```
cpascual: I see that you removed the __subscription_event = threading.Event() initialization...
Is this ok? 
where is it initialized now?

cpascual:
Hi @sergirubio , thanks for the changes. All my original comments have been addressed, but I see that you introduced som extra changes and I do not fully understand them (and the one-line commit messages are not helping).

Can you please explain the reasoning behind the changes?

sergirubio:
I minimized the amount of lines involved to provide more clarity.
The __subscription_event was already created in the __init__ method. As it is used in several methods it would be a risk if it is created twice (overriding its previous state).
For the same reason I added the checks for ..._evt_id not being initialized previously. If the previous event id is overriden it would never be unsubscribed at exit.

cpascual:
thanks @sergirubio for the explanation. It is clear now. Merging

PR #597, opened 2017-10-25
Fix wrong plugin info for TaurusTrend2DDialog

cpascual:
TaurusTrend2DDialog cannot be used in taurusdesigner due to a wrong
module info in the its designer plugin method. Fix it.


cpascual:
To reproduce the problem, create a widget with designer , include a Taurustrend2DDialog on it , save it as `foo.ui` and then try to load the ui:

```
from taurus.external.qt import QtGui, uic
uic.loadUiType("foo.ui") 
```

cpascual:
@taurus-org/integrators Can someone review this and integrate?

reszelaz:
Tested and works. Integrating...

Issue #592, opened 2017-10-20
missing OrderedDict in taurus.external.ordereddict


jkotan:
Hi, 

I've just tried to test taurus on stretch and because of missing workaround for a pyqt bug in Taurus 4.0 (original package for stretch is not usable, i.e. macrogui fails) I decide to go to Taurus 4.1.

Unfortunately, when I try to run sardana I get the following error:

    Traceback (most recent call last):
      File "/usr/bin/Pool", line 11, in <module>
	load_entry_point('sardana==2.2.2', 'console_scripts', 'Pool')()
      File "/usr/lib/python2.7/dist-packages/sardana/tango/pool/__init__.py", line 101, in main
	run(start_time=datetime.datetime.now())
      File "/usr/lib/python2.7/dist-packages/sardana/tango/pool/__init__.py", line 95, in main_pool
	name=SERVER_NAME)
      File "/usr/lib/python2.7/dist-packages/sardana/tango/core/util.py", line 965, in run
	prepare_func(tango_util)
      File "/usr/lib/python2.7/dist-packages/sardana/tango/pool/__init__.py", line 75, in prepare_pool
	from .Pool import PoolClass, Pool
      File "/usr/lib/python2.7/dist-packages/sardana/tango/pool/Pool.py", line 45, in <module>
	from sardana.pool.pool import Pool as POOL
      File "/usr/lib/python2.7/dist-packages/sardana/pool/pool.py", line 56, in <module>
	from sardana.pool.poolcontrollermanager import ControllerManager
      File "/usr/lib/python2.7/dist-packages/sardana/pool/poolcontrollermanager.py", line 40, in <module>
	from taurus.external.ordereddict import OrderedDict
    ImportError: No module named ordereddict
    terminate called after throwing an instance of 'omni_thread_fatal'

Is Taurus 4.0 to 4.1 backward compatible or `lib/taurus/external/ordereddict/__init__.py` is missing?

The following fix

    --- /dev/null
    +++ b/lib/taurus/external/ordereddict/__init__.py
    @@ -0,0 +1,8 @@
    +from __future__ import absolute_import
    +
    +try:
    +    # ordereddict from python 2.7 or from ordereddict installed package?
    +    from collections import OrderedDict
    +except ImportError:
    +    # ordereddict from local import
    +    import os

works for me.

Anyhow, it would be good to update stretch packages for taurus/sardana to working versions.

Bests,
Jan

cmft:
Hi @jkotan ,
[...]

PR #599, opened 2017-10-25
add taurus.external.ordereddict back

jkotan:
Hi,

this is solution for the issue https://github.com/taurus-org/taurus/issues/592.

In taurus from 4.0 - 4.1 it was also removed

According to documentation also

    taurus.qt.qtgui.Q* modules (Qt, QtCore, QtGui, Qwt5,...)
    taurus.qt.qtgui.util.taurusropepatch module
    taurusqt.qtgui.util.genwidget

was removed but I don't know it those changes are revertible (keeping the current interface). 

Bests,
Jan

### DIFF
--- a/lib/taurus/external/ordereddict/__init__.py
+++ b/lib/taurus/external/ordereddict/__init__.py
@@ -26,4 +26,4 @@
 from collections import OrderedDict
 from taurus.core.util import log as __log
 
-__log.deprecated(dep='taurus.external.ordereddict', rel='4.0.4')
+__log.deprecated(dep='taurus.external.ordereddict', rel='4.0.3')
--- a/lib/taurus/external/ordereddict/__init__.py
+++ b/lib/taurus/external/ordereddict/__init__.py
@@ -1,0 +1,8 @@
+from __future__ import absolute_import
+
+try:
+    # ordereddict from python 2.7 or from ordereddict installed package?
+    from collections import OrderedDict
+except ImportError:
+    # ordereddict from local import
+    import os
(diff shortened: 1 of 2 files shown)


lib/taurus/external/ordereddict/__init__.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import absolute_import
+
+try:
```
cpascual: I think that we do not need the try-except or the `__future__` import: In Taurus 4 we depend on python 2.7 which provides collections.OrderedDict

So I would simply do `from collections import OrderedDict`

cpascual: Well... actually I would also add a deprecation warning using `taurus.deprecated`

cpascual: And we also should edit the changelog accordingly

cpascual:
> According to documentation also
> ```
> taurus.qt.qtgui.Q* modules (Qt, QtCore, QtGui, Qwt5,...)
> taurus.qt.qtgui.util.taurusropepatch module
> taurusqt.qtgui.util.genwidget
> ```
> was removed but I don't know it those changes are revertible (keeping the current interface).

AFAIK, all those modules were leftovers from taurus 3. They ought to have been removed in the major transition to Taurus 4.0 but were forgotten in the cleanup. 
Even if *in some strict sense* removing them in 4.0.3->4.1.0 transition goes against semver, I believe that it in practice it is ok to leave them out since the backwards incompatibility belongs to the 3->4 transition and we are just "finishing it". 

Note that this is a different thing from the ordereddict case, where the deprecation was done after the transition to v4 (and therefore we need to re-add it)

cpascual:
I reviewed the latest commits from @jkotan and found them ok.
I did a couple of minor changes myself and I am merging.

PR #605, opened 2017-10-27
Add Delayed subscriber to emitter classes

sergirubio:
Added DelayedSubscriber to emitter classes:

 - DelayedSubscriber is a QThread to retry event subscription once Qt
   is already running.
 - It required to replace _activatePolling() by enablePolling(True) in TangoAttribute
   when tango events are not enabled.
 - This commit adds a polling timer and a loop wait to TaurusEmitterThread,
   in order to manage cpu usage.
 - Reformat __doc__ for emitter classes

### DIFF
--- a/lib/taurus/qt/qtcore/util/emitter.py
+++ b/lib/taurus/qt/qtcore/util/emitter.py
@@ -142,7 +142,8 @@ class TaurusEmitterThread(Qt.QThread):
             def __init__(self, parent = None, designMode = False):
                 ...
                 self.modelsQueue = Queue.Queue()
-                self.modelsThread = TaurusEmitterThread(parent=self,queue=self.modelsQueue,method=modelSetter )
+                self.modelsThread = TaurusEmitterThread(parent=self,
+                        queue=self.modelsQueue,method=modelSetter )
                 ...
             def build_widgets(...):
                 ...
@@ -160,9 +161,10 @@ class TaurusEmitterThread(Qt.QThread):
 
     """
 
-    def __init__(self, parent=None, name='', queue=None, method=None, cursor=None, sleep=5000):
+    def __init__(self, parent=None, name='', queue=None, method=None,
+                 cursor=None, sleep=5000, period_ms=0):
         """
-        Parent most not be None and must be a TaurusGraphicsScene!
+        Parent must be not None and must be a TaurusGraphicsScene!
         """
         Qt.QThread.__init__(self, parent)
         self.name = name
@@ -185,6 +187,12 @@ class TaurusEmitterThread(Qt.QThread):
         # Moved to the end to prevent segfaults ...
         self.emitter.doSomething.connect(self._doSomething)
         self.emitter.somethingDone.connect(self.next)
+        
+        self.period = period_ms
+        
+    def onRefresh(self):
+        self.refreshTimer.setInterval(self.period)
+        self.next()
 
     def getQueue(self):
         if self.queue:
@@ -260,11 +268,18 @@ class TaurusEmitterThread(Qt.QThread):
         return
 
     def run(self):
-        Qt.QApplication.instance().thread().sleep(int(self.timewait / 1000)
-                                                  if self.timewait > 10 else int(self.timewait))  # wait(self.sleep)
+        Qt.QApplication.instance().thread().sleep(
+            int(self.timewait / 1000) if self.timewait > 10 else int(self.timewait))  # wait(self.sleep)
         self.log.info('#' * 80)
         self.log.info('At TaurusEmitterThread.run()')
         self.next()
+        
+        if self.period:
+            self.refreshTimer = Qt.QTimer()
+            Qt.QObject.connect(self.refreshTimer, 
+                            Qt.SIGNAL("timeout()"), self.onRefresh)
+            self.refreshTimer.start(self.period)        
+        
         while True:
             self.log.debug('At TaurusEmitterThread.run() loop.')
             item = self.todo.get(True)
--- a/lib/taurus/qt/qtcore/util/emitter.py
+++ b/lib/taurus/qt/qtcore/util/emitter.py
@@ -39,6 +39,7 @@ from taurus.core.util.singleton import Singleton
 
 from taurus.core.taurusbasetypes import SubscriptionState
 
+
 ###############################################################################
 # Helping methods
 
@@ -215,16 +216,15 @@ class TaurusEmitterThread(Qt.QThread):
         self._done = 0
         # Moved to the end to prevent segfaults ...
         self.emitter.doSomething.connect(self._doSomething)
-        
+
         if not self.refreshTimer:
             self.emitter.somethingDone.connect(self.next)
-        
-        
+
     def onRefresh(self):
         try:
             size = self.getQueue().qsize()
             if size:
-                self.log.info('onRefresh(%s)'%size)
+                self.log.info('onRefresh(%s)' % size)
                 self.next()
             else:
                 self.log.debug('onRefresh()')
@@ -275,8 +275,8 @@ class TaurusEmitterThread(Qt.QThread):
             try:
                 method(*args)
             except:
-                self.log.error('At TaurusEmitterThread._doSomething(%s): \n%s' 
-                    % (map(str, args), traceback.format_exc()))
+                self.log.error('At TaurusEmitterThread._doSomething(%s): \n%s'
+                               % (map(str, args), traceback.format_exc()))
         self.emitter.somethingDone.emit()
         self._done += 1
         return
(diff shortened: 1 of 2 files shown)


lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -23,10 +23,20 @@
...
 
 """
-emitter.py: This module provides a task scheduler used by TaurusGrid and TaurusDevTree widgets
+emitter.py: This module provides a task scheduler used by TaurusGrid and 
+    TaurusDevTree widgets
 """
 
-import Queue
+try:
+    import queue
```
cpascual: Please use only 1 flavour (I suggest the modern one: `queue`) .
According to https://docs.python.org/2/library/queue.html , the `Queue` module was renamed to `queue` **in python 3**

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -23,10 +23,20 @@
...
+    import queue as Queue    
+except:
+    #Python 2.6 backwards compatibility
```
cpascual: Keep in mind that taurus 4 requires python 2.7, so the comment is misleading

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -175,6 +200,13 @@ def __init__(self, parent=None, name='', queue=None, method=None, cursor=None, s
             Qt.Qt.WaitCursor) if cursor is True else cursor
         self._cursor = False
         self.timewait = sleep
+        self.polling = polling
+        self.loopwait = int(loopwait*1e-3)
```
cpascual: Are you sure this is ok?
The default value of loopwait is 5. With this cast, the default value is going to be 0. Maybe you wanted to multiply by 1e3?

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -184,7 +216,22 @@ def __init__(self, parent=None, name='', queue=None, method=None, cursor=None, s
...
         self.emitter.doSomething.connect(self._doSomething)
-        self.emitter.somethingDone.connect(self.next)
+        
+        if not self.refreshTimer:
+            self.emitter.somethingDone.connect(self.next)
+        
+        
+    def onRefresh(self):
+        try:
+            self.refreshTimer.setInterval(self.polling)
```
cpascual: This seems **very** ugly to me. You are resetting the interval of the timer **on each callback from the timer**

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -196,7 +243,8 @@ def getQueue(self):
 
     def getDone(self):
         """ Returns % of done tasks in 0-1 range """
-        return self._done / (self._done + self.getQueue().qsize()) if self._done else 0.
+        return self._done / (self._done + self.getQueue().qsize()) \
```
cpascual: Please use a more readable if-else construct.
(PEP8 is not only about putting inserting line breaks at the middle of over-complex constructs) ;)

cpascual:
Some general comments regarding PEP8:
- please do not mix unrelated PEP8 changes with the "actual" changes. It makes reviewing a lot more difficult. It is too late for this PR, but please try to do the PEP8-only changes in independent commits
- Please have a look at PEP8 regarding the line breaks. In particular this about avoiding bacslash for line breaking:
  > The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

sergirubio:
Ok, I did all the aforementioned changes and tested it against lab devices. 
Note that "Queue" module is still imported from other files in taurus.

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -130,7 +142,7 @@ class TaurusEmitterThread(Qt.QThread):
     .. code-block:: python
 
         #Applying TaurusEmitterThread to an existing class:
-        import Queue
+        import queue
```
cpascual: maybe this should be `from queue import Queue` ?

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -252,19 +303,24 @@ def next(self):
...
         except:
             self.log.warning(traceback.format_exc())
         return
 
     def run(self):
-        Qt.QApplication.instance().thread().sleep(int(self.timewait / 1000)
-                                                  if self.timewait > 10 else int(self.timewait))  # wait(self.sleep)
+        Qt.QApplication.instance().thread().sleep(
+            int(self.timewait / 1000) if self.timewait > 10 
+                                else int(self.timewait))
```
cpascual: It is not critical, but the following would be a lot more readable (and in PEP8 spirit):

```python
_sleep = self.timewait
if self.timewait > 10:
    _sleep /= 1000
Qt.QApplication.instance().thread().sleep(int(_sleep))
```
(it is 4 lines instead of 3, but it is much easier to follow)

sergirubio: I removed the entire condition, it was an old tau-compatibility check that is not used anymore.

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -206,7 +252,7 @@ def clear(self):
             self._done += 1
 
     def purge(obj):
```
cpascual: I am confused here: the purge method does not define `self`, yet it uses it.

I've searched and I  do not see the purge method being called at all in Taurus, or it being part of an Qt API,... so I am tempted to simply remove the `purge` method, **but**...

I see the same thing happening in other classes in this same module and this PR actually does some changes in the purge method and uses `self` there, so I am worried that I am missing something... 

Can anyone clarify if this is just buggy leftover code or if somehow it makes sense?

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -275,33 +329,110 @@ def run(self):
...
+            if self.loopwait:
+                self.msleep(self.loopwait)
+                # End of while
         self.log.info(
             '#' * 80 + '\nOut of TaurusEmitterThread.run()' + '\n' + '#' * 80)
         # End of Thread
 
 
-class SingletonWorker():  # Qt.QObject):
+class DelayedSubscriber(Logger):
```
cpascual: Please document the methods of this new class.
At least all *public* methods should be documented
Private/protected methods may not be documented if they are obvious enough, but it does not harm
Also, consider minimizing the amount of public API in the class (i.e. make members/methods private unless there is some reason to believe they may be of interest externally (this way we do notcommit into maintainng a large API and we facilitate future refactorings)

cpascual: Also, consider minimizing the amount of public API in the class (i.e. make members/methods private unless there is some reason to believe they may be of interest externally (this way we do notcommit into maintainng a large API and we facilitate future refactorings)

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -275,33 +333,110 @@ def run(self):
...
+
+    def addModelObj(self, modelObj):
+        parent = modelObj.getParentObj()
+        if parent:
+            proxy = parent.getDeviceProxy()
+            if not proxy:
+                # self.debug('addModelObj(%s), proxy not available'%modelObj)
+                return
+        self._modelsQueue.put((modelObj._subscribeConfEvents,))
+        modelObj.__subscription_state = SubscriptionState.PendingSubscribe
```
cpascual: This is probably wrong code (useless at least) due to name mangling preventing to actually modify the *protected* `__subscription_state` of `modelObj`

sergirubio:
I did merge against the wrong origin ... there's a way to undoing that?

cpascual:
I already fixed it for you. You can now do the merge against taurus-org/taurus develop

lib/taurus/qt/qtcore/util/emitter.py
```diff
@@ -275,33 +331,108 @@ def run(self):
...
+
+    def addModelObj(self, modelObj):
+        parent = modelObj.getParentObj()
+        if parent:
+            proxy = parent.getDeviceProxy()
+            if not proxy:
+                self.debug('addModelObj(%s), proxy not available' % modelObj)
+                return
+
+        self._modelsQueue.put((modelObj._subscribeConfEvents,))
```
cpascual: We should not access a private method from outside its class

cpascual:
I only did a code-review, but I did not actually test the new class, since it requires quite a lot of extra code to test.
I think it is safe to merge because it does not affect any existing functionality but I do cannot guarantee that the DelayedSubscriber works as intended.

@sergirubio: it would be extremely appreciated if you provided a unit test for DelayedSubscriber, in order to facilitate the maintainability of this class (but in order not to block I will merge this PR and expect the unit test to come as another PR).

PR #608, opened 2017-11-03
Remove PyTango dependency in taurusform.py

sergirubio:


cpascual:
Please @sergirubio , add some justification in your PRs and commits. For example, in this case, we, integrators only have a one-line sentence to figure out the rationale behind your change.

For example, your commit message could have been:

```
Remove PyTango dependency in taurusform.py

TaurusCommandForm uses PyTango without importing it, causing an exception. 
Remove PyTango usage.
```

This would have helped me to
a) quickly understand the context of the problem
b) understand your motivations for the change

cpascual:
Now, regarding the proposed solution: I do not think it is the best approach because it is assuming that `PyTango.CmdArgType.DevVoid` evaluates (and will always evaluate) as `False`. This is probably going to be the case, but nevertheless it is an antipattern (relying on the value of an enum instead of the enum member).

Note that since the problematic code is in a method  of `TaurusCommandForm` (which is already Tango-centric), we could simply do a local import of `PyTango` before the problematic line.

sergirubio:
I modified the type check using taurus.core.tango

cpascual:
Seems good to me. Thanks!

PR #609, opened 2017-11-03
Migrate TaurusGrid.itemClicked signal to API 2 (bugfix)

sergirubio:
TaurusValue has no itemClicked signal, apart of that there was conflicts between existing methods and signal names.

### DIFF
--- a/lib/taurus/qt/qtgui/table/taurusgrid.py
+++ b/lib/taurus/qt/qtgui/table/taurusgrid.py
@@ -887,7 +887,7 @@ class TaurusGrid(QtGui.QFrame, TaurusBaseWidget):
             widgets_row = []
             for cell in row:
                 cell_frame = self.create_frame_with_gridlayout()
-                cell_frame.itemClicked.connect(self.itemClickedHook)
+                cell_frame.itemClicked.connect(self.onItemClicked)
                 count = 0
                 for synoptic in sorted(cell):
                     self.debug("processing synoptic %s" % synoptic)
@@ -910,7 +910,7 @@ class TaurusGrid(QtGui.QFrame, TaurusBaseWidget):
             widgets_matrix.append(widgets_row)
         return widgets_matrix
 
-    def itemClickedHook(self, item_name):
+    def onItemClicked(self, item_name):
         self.trace('In TaurusGrid.itemClicked(%s)' % item_name)
         self.setItemSelected(item_name)
         self.itemClicked.emit(str(item_name))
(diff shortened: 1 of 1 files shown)


lib/taurus/qt/qtgui/table/taurusgrid.py
```diff
@@ -894,23 +906,11 @@ def build_widgets(self, values, show_labels=False, width=240, height=20, value_w
...
-                cell_frame.mousePressEvent = partial(
-                    mousePressEvent, obj=cell_frame)
-                cell_frame.itemClicked.connect(self.itemClicked)
-
                 widgets_row.append(cell_frame)
             widgets_matrix.append(widgets_row)
         return widgets_matrix
 
-    def itemClicked(self, item_name):
+    def itemClickedHook(self, item_name):
```
cpascual: Not important, but for consistency with other parts of Taurus, could you consider renaming this to `onItemClicked`?
...or even better, make this private by calling it `_onItemClicked`

lib/taurus/qt/qtgui/table/taurusgrid.py
```diff
@@ -194,6 +194,18 @@ def get_readwrite_models(expressions, limit=1000):
     models = models[:limit]
     return models
 
+class TaurusGridCell(Qt.QFrame):
```
cpascual: Since this class seems to be a very private helper for `TaurusGrid`, I would explicitly mark it as private by calling it `_TaurusGridCell` (this way we avoid having to worry about API compatibility if we refactor in the future)

cpascual:
Apart from 2 minor comments on naming, the code changes seem ok to me.

cpascual:
Hi @sergirubio , do you also agree with the rename of `TaurusGridCell` --> `_TaurusGridCell `?

sergirubio:
I would keep it public, it may be used from other synoptic-like widgets.

lib/taurus/qt/qtgui/table/taurusgrid.py
```diff
@@ -194,6 +194,18 @@ def get_readwrite_models(expressions, limit=1000):
...
+
+    itemClicked = Qt.pyqtSignal('QString')
+    
+    # Done in this way as TaurusValue.mousePressEvent is never called
```
cpascual: Is the reference to TaurusValue in this comment ok?

cpascual:
>I would keep it public, it may be used from other synoptic-like widgets.

Ok, but... is the mouse event hack in that class something that we really want in a public class?
(it looks a bit ugly and "ad-hoc" to me)

And in any case , if you insit in making this class public, then you should:
a) list it in the `__all__` variable in taurusgrid.py so that it is accessible and 
b) document the class (and its only method)

sergirubio:
ok

PR #616, opened 2017-11-15
Remove unwanted debug log

reszelaz:
This debug log was introduced in b42b98ca0 and I think it is not necessary.


cpascual:
merging, thanks!

PR #617, opened 2017-11-16
Cast to bool the value in _TaurusLedController

cmft:
TaurusLed in taurus 3 allows to use not boolean models, if its
values are '0' or  '1'.

e.g. eval:1 or eval:0

Improve the behaviour of the TaurusLed casting the value to bool for
non boolean models.


cpascual:
Merged, Thanks!

PR #618, opened 2017-11-16
Catch disconnect exception in TaurusTrend

sergirubio:
Whenever calling "taurustrend -a <attribute_name>" and changing the X scale, the following exception appeared and the plot wasn't updated:

```
 File "/homelocal/sicilia/local/taurus/qt/qtgui/plot/taurustrend.py", line 1619, in _scaleChangeWarning 
   self.showArchivingWarning() 
 File "/homelocal/sicilia/local/taurus/qt/qtgui/plot/taurustrend.py", line 1629, in showArchivingWarning 
   self.axisWidget(self.xBottom).scaleDivChanged.disconnect(self._scaleChangeWarning) 
TypeError: disconnect() failed between 'scaleDivChanged' and 'unislot'
```

Probably it failed because the signal was not connected. It is a minor issue and there's no method to check the connection. I just added a try/except to print a warning and proceed to plot the data.

### DIFF
--- a/lib/taurus/qt/qtgui/plot/taurustrend.py
+++ b/lib/taurus/qt/qtgui/plot/taurustrend.py
@@ -1626,7 +1626,12 @@ class TaurusTrend(TaurusPlot):
         # stop the scale change notification temporally (to avoid duplicate
         # warnings)
         self.setUseArchiving(False)
-        self.axisWidget(self.xBottom).scaleDivChanged.disconnect(self._scaleChangeWarning)
+        try:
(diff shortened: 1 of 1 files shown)


cpascual:
> Probably it failed because the signal was not connected. It is a minor issue and there's no method to check the connection. 

Yes. This is a side-efect of changing to new-style signals. Old-style signals were ok if disconnected without being previously connected but new-style signals are not.

>I just added a try/except to print a warning and proceed to plot the data.

Fine for me

PR #623, opened 2017-11-17
Solve Command+ComboBox bug, setParameters expected string not int

sergirubio:
QComboBox.currentIndexChanged is an overloaded signal, it can send either int or QString

The original code in Taurus 3  was:

self.connect(pwidget, Qt.SIGNAL('currentIndexChanged (const QString&)'),button.setParameters)

But when changing to API2 the "QString" modifier overloading was lost, it caused exceptions in setParameters as it expected string and received int instead. 

I tested the patch using Vacca and it works


cpascual:
Yes, we have seen this same issue in other cases. Thanks for the patch!

PR #629, opened 2017-11-21
Improve display widget Controllers

### LINKED ISSUE #626: TaurusLabel: allow to represent *tango* states as background
Setting the model of a TaurusLabel to a DevState attribute of a tango device, with the bgRole configured to represent the state, it doesn't change to blue in case of a running state. But very worst case when the state is fault, because then it shows a text "FAULT" in a green background.

If I understood correctly, it seems taurus4 is only supporting 3 TaurusDevState (Ready, NotReady and Unknown). Like there is a TangoAttribute that inherits from TaurusAttribute and give the details when it point to a tango attribute, there have to be a TangoDevState that inherits from TaurusDevState and encapsulates this behaviour.

cpascual:
Extend bgRole API of TaurusController to allow passing the name of a property of the model object as the bgRole.

This allows e.g. to simulate the behaviour of Taurus3 state labels by setting the model to a state and the bgRole to "rvalue":

```python
import sys
app = TaurusApplication()
w = TaurusLabel()
w.setModel( 'sys/tg_test/1/state')
w.setBgRole('rvalue')
w.show()
sys.exit(app.exec_())
```

<s>This Fixes #626</s> Update: actually, #626 does not really need this to be closed

Also, fix a related bug with state leds.

Finally, remove the state led of taurusdevicepanels since it is no longer needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ develop branch) won't be reflected in this file.
 - Option to ignore outdated Tango events (#559)
 - Travis-built docs (not yet replacing the RTD ones) (#572)
 - TaurusLed now supports non-boolean attributes (#617)
+- Support for arbitrary bgRole in labels (#629)
 
 ### Changed
 - taurus.qt widgets can now be used without installing PyTango (#590)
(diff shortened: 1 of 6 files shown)


cpascual:
@taurus-org/integrators: This is ready for being integrated

reszelaz:
I've reviewed the changes and make tests. Everything is ok. Merging it..

PR #631, opened 2017-11-21
Avoid to import queue from python3

cmft:
The Queue module has been renamed to queue in Python 3.
The emitter module imports the "Queue" module from
python 3, since taurus uses python 2.7 this module
should be imported from python 2 version.

Change the import.

### DIFF
--- a/lib/taurus/qt/qtcore/util/emitter.py
+++ b/lib/taurus/qt/qtcore/util/emitter.py
@@ -27,7 +27,7 @@ emitter.py: This module provides a task scheduler used by TaurusGrid and
     TaurusDevTree widgets
 """
 
-from queue import Queue, Empty
+from Queue import Queue, Empty
 import traceback
 from functools import partial
 from collections import Iterable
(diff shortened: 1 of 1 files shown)


cpascual:
Have you experienced any problem?

This change was introduced in #605 , and in principle, things work fine...
I wonder if `Queue` is somehow providing a `queue` alias in python2.7 (if that is the case, I would leave it as it is)

Anyway, note that the one you change is not the only import of `queue`. For consistency, all should be changed (if we go ahead with this PR).

cpascual:
> Have you experienced any problem?

 I answer myself: I just tested using a fresh python2.7 virtualenv and `queue` cannot be imported.

I suspect that we did not see this in #605 because on my machine I've got both python2 and python3 installed  (or maybe the python debian package provides the alias?)

in any case, I agree in changing back to `Queue`, but all imports need changing

cpascual:
I did the search and found that (again) I was wrong. There are no more imports of `queue`, only a mention in a docstring (which I just fixed).

So I am merging

PR #632, opened 2017-11-22
Add import ascii file option to taurusplot

sergirubio:
Some users asked me for an script to plot archiving exports using taurus.

It was so easy to do that I think it's worth adding the option directly on taurus.

The idea is to generate raw data from archiving and/or matlab:
`archiving2csv --noheader --noepoch <model1> <model2> <date1> <date2> /tmp/file.csv`

the generated file can then be opened by taurusplot just doing:
`taurusplot -x t --import=/tmp/file.csv`


cpascual:
Thanks for the feature.
I merged it with one change: I renamed the option to "--import-ascii" and removed the "--import" option.

cpascual:
Note: this is option can also be used with the file resulting from using the "Export data to ASCII" action from "Export & Print" in the context menu.

PR #636, opened 2017-11-24
(doc) Add tip about the new panel dlg UI

cpascual:
Add tip about the "other..." button in the "new panel" dialog


cpascual:
Note: the resulting doc can be seen in: 
https://taurus-org.github.io/taurus-doc/v-doc-tip/users/ui/taurusgui.html#creating-custom-panels

PR #640, opened 2017-12-12
Add full model name to attribute description

cpascual:
The attribute description is used by widgets to generate the tooltip.
Add full model name info to the description object


cmft:
It is ready to be integrated!.

PR #644, opened 2017-12-13
Convert gui_scripts to console_scripts (Fixes #541)

### LINKED ISSUE #541: windows taurus4: launcher console is closed when application is terminated
win7 taurus4: launcher console is closed when application is terminated. This can be considered a bug, because when problems appear with taurus applications, the error in the console is not visible, because the console is automatically closed, and it is then difficult to track the bugs.

As it is now, the help of the taurus applications cannot be seen. For instance, by doing 'taurusplot --help', the console closes too fast, and the help cannot be read.

cpascual:
Launchers do not show output in Windows if they are implemented
as gui_scripts. Change them to console_scripts.

Fixes #541

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,10 +81,6 @@ extras_require = {
 
 console_scripts = [
     'taurustestsuite = taurus.test.testsuite:main',
-    # TODO: taurusdoc,
-]
-
-gui_scripts = [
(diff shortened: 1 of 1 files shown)


cpascual:
Merging after mrosanes approved review

PR #646, opened 2017-12-14
Fix: Taurusform does not allow to change the label

### LINKED ISSUE #641: Taurusform does not allow to change the label
Taurusform does not allow to change the label. The bug seems to be introduced by #496

cmft:
The PR #496 added a regresion in the taurusform change label feature.

Fix it by:

- In DefaultLabelWidget class:
  - revert the changes in _BCK_COMPAT_TAGS
  - fix getDisplayValue

- Fix taurusValue onChangeLabelConfig method.

- Reimplement taurusLabel setText to use setPemanentText.
- Fix displayValue to allow format strings

Fix #641

### DIFF
--- a/lib/taurus/qt/qtgui/display/tauruslabel.py
+++ b/lib/taurus/qt/qtgui/display/tauruslabel.py
@@ -457,14 +457,22 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
 
     def displayValue(self, v):
         """Reimplementation of displayValue for TaurusLabel"""
-        if self._permanentText is not None:
-            value = self._permanentText
-        else:
+        if self._permanentText is None:
             value = TaurusBaseWidget.displayValue(self, v)
+        else:
+            value = self._permanentText
 
         attr = self.getModelObj()
         dev = attr.getParent()
-        return value.format(dev=dev, attr=attr)
+
+        try:
+            v = value.format(dev=dev, attr=attr)
+        except Exception as e:
+            self.warning(
+                "Error formatting display (%r). Reverting to raw string", e)
+            v = value
+
+        return v
 
     @classmethod
     def getQtDesignerPluginInfo(cls):
--- a/lib/taurus/qt/qtgui/display/tauruslabel.py
+++ b/lib/taurus/qt/qtgui/display/tauruslabel.py
@@ -252,7 +252,7 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
 
         # register configurable properties
         self.registerConfigProperty(
-            self.getPermanentText, self.setPermanentText, "permanentText"
+            self.getPermanentText, self._setPermanentText, "permanentText"
             )
 
     def _calculate_controller_class(self):
@@ -418,9 +418,10 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
     def getPermanentText(self):
         return self._permanentText
 
-    def setPermanentText(self, text):
-        self.setText_(text)
+    def _setPermanentText(self, text):
         self._permanentText = text
+        if text is not None:
+            self.setText_(text)
 
     def setText_(self, text):
         """Method to expose QLabel.setText"""
@@ -428,7 +429,7 @@ class TaurusLabel(Qt.QLabel, TaurusBaseWidget):
 
     def setText(self, text):
         """Reimplementation of setText to set permanentText"""
-        self.setPermanentText(text)
+        self._setPermanentText(text)
 
     def setAutoTrim(self, trim):
         self._autoTrim = trim
--- a/lib/taurus/qt/qtgui/display/tauruslabel.py
+++ b/lib/taurus/qt/qtgui/display/tauruslabel.py
@@ -98,7 +98,7 @@ class TaurusLabelController(TaurusBaseController):
         self._trimmedText = self._shouldTrim(label, text)
         if self._trimmedText:
             text = "<a href='...'>...</a>"
-        label.setText(text)
+        label.setText_(text)
 
     def _shouldTrim(self, label, text):
         if not label.autoTrim:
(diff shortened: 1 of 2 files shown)


lib/taurus/qt/qtgui/display/tauruslabel.py
```diff
@@ -419,9 +419,17 @@ def getPermanentText(self):
         return self._permanentText
 
     def setPermanentText(self, text):
```
cpascual: I think that setPermanentText() should be renamed to setText()
Note that setPermanentText has not been in any released version of Taurus (and therefore we should not worry about compatibility)

lib/taurus/qt/qtgui/display/tauruslabel.py
```diff
@@ -448,9 +456,15 @@ def resetAutoTrim(self):
         self.setAutoTrim(self.DefaultAutoTrim)
 
     def displayValue(self, v):
+        """Reimplementation of displayValue for TaurusLabel"""
         if self._permanentText is not None:
```
cpascual: invert logic. Use `is None`

lib/taurus/qt/qtgui/display/tauruslabel.py
```diff
@@ -448,9 +456,15 @@ def resetAutoTrim(self):
...
-            return self._permanentText
-        return TaurusBaseWidget.displayValue(self, v)
+            value = self._permanentText
+        else:
+            value = TaurusBaseWidget.displayValue(self, v)
+
+        attr = self.getModelObj()
+        dev = attr.getParent()
+        return value.format(dev=dev, attr=attr)
```
cpascual: protect this against exceptions in format(). Something like:
```python
try:
    v = value.format((dev=dev, attr=attr))
except Exception as e:
    self.warning("Error formatting display (%r). Reverting to raw string", e)
    v = value
return v
```

lib/taurus/qt/qtgui/panel/taurusvalue.py
```diff
@@ -1233,19 +1236,21 @@ def setLabelConfig(self, config):
         """
         # backwards compatibility: this method used to work for setting
         # an arbitrary text to the label widget
+        self._labelConfig = config
         try:
```
cpascual: IMHO, setLabelConfig needs to be rewritten. The problems I see:
- It should do what the docstrings says: set a fragment, not an arbitrary formar str
-  the first try-except makes no sense as it is now (unless I am misunderstanding something)...

Perhaps setLabelConfig should be deprecated and let only for the old  purpose with a warning, and only `setText()` / `setText_()` should be used internally

lib/taurus/qt/qtgui/panel/taurusvalue.py
```diff
@@ -91,7 +91,7 @@ def setModel(self, model):
...
             config = self.taurusValueBuddy().getLabelConfig()
         except Exception:
-            config = 'label'
+            config = '{attr.label}'
```
cpascual: see https://github.com/taurus-org/taurus/pull/646/files#r156952262

lib/taurus/qt/qtgui/display/tauruslabel.py
cmft: `permanentText` is a property with setter and getter `getPermanentText and setPermanentText` 
Do you propose to change this API? or just change setPermanentText by setText?

lib/taurus/qt/qtgui/display/tauruslabel.py
cmft: Ok

lib/taurus/qt/qtgui/display/tauruslabel.py
cmft: Ok, but it is not new code.

lib/taurus/qt/qtgui/panel/taurusvalue.py
cmft: `self._labelConfig` is used to keep the current configuration, and selected it in onChangeLabelConfig 
and should be move before the comment.

It was done like this in #496, IMO we could change the docstring to accept format string and we could simplify the check

> Perhaps setLabelConfig should be deprecated and left only for providing bck-compat, and only setText() / setText_() should be used internally. The same for {get,reset}LabelConfig.
Maybe, but I could be a "big" API change

lib/taurus/qt/qtgui/panel/taurusvalue.py
cmft: I think, it is allowed for labelConfig

lib/taurus/qt/qtgui/display/tauruslabel.py
cpascual: good point... actually if we want to provide allow the config property to work, we need to keep it... 
So I change my proposal to renaming `setPermanentText` to `_setPermanentText` (so that there is only 1 public accessor).

lib/taurus/qt/qtgui/panel/taurusvalue.py
cpascual: I personally would prefer to deprecate {get,set,reset}LabelConfig (to simplify an already messy API)

lib/taurus/qt/qtgui/display/tauruslabel.py
cmft: Done!

reszelaz:
Tested with Sardana widgets. Below what is displayed in the label text :

taurus version | motor | counter | ior
------------------ | ------- | ---------- | -----
support-3.x | dev name | dev name | dev name
4.1.1 | dev name | dev name | dev name
develop | dev name | attr name (Value) | dev name
this PR | dev name | string <dev_alias> | dev nane

So there is something what is broken for the counter widget since the last release.

cmft:
@reszelaz  I fixed an introduced bug. Could you test it again?

reszelaz:
Yes, it works as expected now. Thanks. Merging it..

PR #650, opened 2017-12-19
TaurusValueLineEdit Bug

sergirubio:
This bug affects TaurusValueLineEdit widget when events are disabled or not available. The widget is meant to keep the last value introduced by user and change to "blue" if the value is modified by other client.

But, in the case when the setModel() methods ends before the Polling timer has started the value of the widget is set to None ... and never updated afterwards, not even changing to "blue" when required.

To solve it, I overwrite the existing value if it was None and there are new pending operations (value has changed).

No operation is applied as a result of this change and we have tested with C.Falcón that it doesn't affect existing taurusforms.

### DIFF
--- a/lib/taurus/qt/qtgui/input/tauruslineedit.py
+++ b/lib/taurus/qt/qtgui/input/tauruslineedit.py
@@ -237,7 +237,7 @@ class TaurusValueLineEdit(Qt.QLineEdit, TaurusBaseWritableWidget):
             elif model_type == DataType.Bytes:
                 return bytes(text)
             else:
-                raise TypeError('Unsupported model type "%s"', model_type)
+                raise TypeError('Unsupported model type "%s"'%model_type)
         except Exception, e:
             self.warning('Cannot return value for "%s". Reason: %r', text, e)
             return None
--- a/lib/taurus/qt/qtgui/input/tauruslineedit.py
+++ b/lib/taurus/qt/qtgui/input/tauruslineedit.py
@@ -122,6 +122,12 @@ class TaurusValueLineEdit(Qt.QLineEdit, TaurusBaseWritableWidget):
         TaurusBaseWritableWidget.updateStyle(self)
 
         value = self.getValue()
+        if value is None and self.hasPendingOperations():
+            try:
+                value = self.getModelValueObj().wvalue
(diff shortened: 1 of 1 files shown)


cpascual:
Merged, thanks!

cpascual:
The changes from this PR are problematic (see #667).
Also, while looking for a solution to #667 , I noticed that probably the description of this issue was not completely accurate: at least I can not exactly reproduce the reported issue.

What I can reproduce however is the fact that a TaurusValueLineEdit whose model is numeric and whose text stars with `0` is, for some reason, interpreted as displaying `0`, regardless of the text (e.g., `0123` is interpreted as `0`.

This may lead to some symptons such as those described in the [original description](https://github.com/taurus-org/taurus/pull/650#issue-283282985).

@sergirubio , could you please check if issue #650 has been reintroduced when using code from #673?
And if so, can you please confirm if the issue #650 is actually a problem with leading zeroes?
(If that is not the case, please provide detailed instructions to properly reproduce #650)

sergirubio:
This bug you mention regarding values starting with 0 ... does it occurs with master taurus code or with this PR only?

cpascual:
On Wednesday, January 24, 2018 12:26:41 AM CET sergirubio wrote:
> This bug you mention regarding values starting with 0 ... does it occurs
> with master taurus code or with this PR only?

I see it in 4.0.3 and 4.1.1 as well

-- 
+----------------------------------------------------+
 Carlos Pascual Izarra
 Scientific Software Coordinator
 Computing Division
 ALBA Synchrotron  [http://www.albasynchrotron.es]
 Carrer de la Llum 2-26
 E-08290 Cerdanyola del Valles (Barcelona), Spain
 E-mail: cpascual@cells.es
 Phone: +34 93 592 4428
+----------------------------------------------------+

cpascual:
> What I can reproduce however is the fact that a TaurusValueLineEdit whose model is numeric and whose text stars with 0 is, for some reason, interpreted as displaying 0, regardless of the text (e.g., 0123 is interpreted as 0).

I tracked this and it seems a problem in pint, which I reported in https://github.com/hgrecco/pint/issues/614 and I included a workaround in #673 (5e7b638)

PR #655, opened 2017-12-21
Improve TangoSchemeTest device server

cmft:
Add two new attributes,  "float_scalar_poll" and "float_scalar_evt",
the first one has been configured to emit Period events
(the tango polling has been enabled), and the second one
has been configured to emit Change events.

There are two new commands to configure the periods of these events:
"SetEventPeriod" to set fixed change events, and "ChangePollingPeriod"
to change the polling period of "float_scalar_poll" attribute.


cpascual:
Looks good to me. The only comment is that I usually prefer to have something more deterministic than a random value in a device that is primarily intended for unit tests...
So I am going to propose a small change in which instead of returning random values, I alternate between a positive and a negative value on each event

PR #661, opened 2018-01-09
Fix for issue #660: added bolded text when displayed value differs fr…

### LINKED ISSUE #660: TaurusValueComboBox does not change color
TaurusValueComboBox does not change color when displayed value differs from applied value.

Other widgets, for example TaurusValueLineEdit, have this nice feature that the foreground color in such cases changes to blue, like this:
screenshot

Is there any simple trick to make this happen on TaurusValueComboBox or is this feature missing there?

I'm using TaurusQt 4.1.1

mkomel:
Make  TaurusValueComboBox use bold font when having pending operations
Fixes #660


cpascual:
I did a couple of minor changes and I merged it.

Thanks @mkomel !

PR #662, opened 2018-01-09
Treat "No unit" as a synonim of PyTango.constants.UnitNotSpec

### LINKED ISSUE #584: taurus.core.tango: "No unit" to be considered unitless
The solution to set unit = " " is not practical as any dynamic attribute created or renamed will loose this config on change. It also creates unnecessary entries into the database that will slow down all queries done to attribute properties values (e.g. to load memorized values from sardana).

To solve that, the taurus.core.tango module should be aware the already known Tango defines for those properties that are "not initialized":

"no unit"
"not defined"
"not specified"
"nada"

This already known defines should be understood by the taurus.core.tango module; for example to recognize a "no unit" attribute as unitless.

cpascual:
Fix #584

To be coherent with the current state of things in Tango and PyTango, I think that both "No unit"  and `PyTango.constants.UnitNotSpec` should be treated equally by taurus.

Note that this may change [if Tango decides to differentiate](https://github.com/tango-controls/TangoTickets/issues/8)

Currently Taurus is **silently** interpreting `PyTango.constants.UnitNotSpec` as unitless. i.e., tango is **not** giving a warning when the unit is `PyTango.constants.UnitNotSpec`, and is simply treating it as unitless. The warnings that we have observed are actually due to the fact that PyTango is returning "No unit" instead of `PyTango.constants.UnitNotSpec`. 

As I said above, I think that both should be treated in the same way, at least while (Py)Tango does not diferentiate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ develop branch) won't be reflected in this file.
 - Model info in widget tooltips (#640)
 
 ### Changed
+- Treat unit="No unit" as unitless in Tango attributes (#662)
 - taurus.qt widgets can now be used without installing PyTango (#590)
 - Tango model name validators now always return FQDN instead of PQDN
   for the tango host (#488)
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -1027,7 +1027,10 @@ class TangoAttribute(TaurusAttribute):
     def _unit_from_tango(self, unit):
         # silently treat unit-not-defined as unitless
         # TODO: consider logging that unit-not-defined is treated as unitless
-        # TODO: See https://github.com/taurus-org/taurus/issues/584
+        # TODO: See https://github.com/taurus-org/taurus/issues/584 and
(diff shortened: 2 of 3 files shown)


reszelaz:
I think that this PR would not be necessary in the scenarios where the server/database side and the client side use the same Tango version. My conclusion comes from the fact that the Tango versions < 9 defines: `PyTango.constants.UnitNotSpec` as `No unit`.

Since we may have mixed environments the extra condition proposed in this PR and comparing to `No unit` is necessary.

For example in a scenario where the server and the database runs on Tango 7 and the client runs on Tango 9, the attribute configuration returns `No unit` if the unit is not defined (no record in the database) and the `PyTango.constants.UnitNotSpec` constant holds an empty string.

Important to know, in Tango9, "No unit" string is no more equivalent to "Not specified". It could look like this in Jive - it gets converted to None immediatelly after pressing enter, but it actually creates an entry in the database with unit set to "No unit".

Hence I'm merging this PR until we get know the Tango point of view about the units. I will just add an extra comment in the code that this extra check is for the Tango < 9.

PR #663, opened 2018-01-10
Fix infinite recursion problem in TaurusValueCombobox

### LINKED ISSUE #651: TaurusValueComboBox repeatedly invokes writeValue
I have noticed that TangoValueComboBox recursively invokes writeValue (at least in TaurusQT 4.1.1). This can be observed in device servers receiving write calls in fast succession.
The problem seems to be in taurusbase.py, class TaurusBaseWritableWidget, function writeValue, lines:

    if applied:
        self.applied.emit()

If I comment those two lines out, it works ok (write operation gets invoked only once, as expected).
Essentially, the effect of self.applied.emit() here is that writeValue gets recursively invoked, until the following error occurs:
    RuntimeError: maximum recursion depth exceeded while calling a Python object.

cpascual:
TaurusValueCombobox connects its own applied signal to the writeValue
method, which in turn emits the applied signal, causing an infinite
loop. Break the loop by not doing the connection (and calling
writeValue directly when needed).

Fixes #651


reszelaz:
Everything is fine. Recheck with snipped provided in #651. Code looks ok. Merging it..

PR #669, opened 2018-01-18
Spinbox was showing units even if the model use fragments to display only the magnitude

srgblnch:
I've seen that a TaurusValueSpinBox was still showing the units in widgets where the model say, in the fragment, that only the magnitude was to be shown.

This pull request contains 2 commits. The first one was changes in the main function at the end of the file, for testing purposes (as well as a pep8 review). Then, like the main in the tauruslabel one can pass some attributes as arguments.

The commit with the solution itself is the second. There is a single change in the TaurusValueLineEdit.setValue(self, v). It was using self.displayValue(v) when the method that manages with the fragments is getDisplayValue.

/Sergi.

### DIFF
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -2061,6 +2061,30 @@ class TaurusBaseWritableWidget(TaurusBaseWidget):
         '''
         return []
 
+    def getDisplayValue(self, cache=True, fragmentName=None):
+        """Returns a string representation of the model value associated with
+        this component.
+
+        :param cache: (bool) (ignored, just for bck-compat).
+        :param fragmentName: (str or None) the returned value will correspond
+                        to the given fragmentName. If None passed,
+                         self.modelFragmentName will be used, and if None is
+                         set, the defaultFragmentName of the model will be used
+                         instead.
+
+        :return: (str) a string representation of the model value.
+        """
+        # @fixme: Even the widgets inheriting from this class interacts with
+        #         writable models (then the fragmentName by default that one
+        #         expects is 'wvalue' and not 'rvalue') it has been considered
+        #         that perhaps this could be considered an API modification
+        #         instead of a bugfix.
+        #         The bugfix impact has been bounded only within the
+        #         TaurusValueLineEdit.
+        return super(TaurusBaseWritableWidget,
+                     self).getDisplayValue(cache=cache,
+                                           fragmentName=fragmentName)
+
     def getValue(self):
         '''
         This method must be implemented in derived classes to return
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -2081,9 +2081,8 @@ class TaurusBaseWritableWidget(TaurusBaseWidget):
         #         instead of a bugfix.
         #         The bugfix impact has been bounded only within the
         #         TaurusValueLineEdit.
-        return super(TaurusBaseWritableWidget,
-                     self).getDisplayValue(cache=cache,
-                                           fragmentName=fragmentName)
+        return TaurusBaseWidget.getDisplayValue(self, cache=cache,
+                                                fragmentName=fragmentName)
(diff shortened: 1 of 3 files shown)


cpascual:
The proposed implementation has one problem: now by default the shown value is rvalue (while in this widget it shoud be wvalue).

This can be checked easily with an attribute such as `sys/tg_test/1/short_scalar` whose wvalue and rvalue are not the same:

```python

if __name__ == '__main__':
    import sys
    app = TaurusApplication()

    w = TaurusValueLineEdit()

    model = 'sys/tg_test/1/short_scalar'
    w.setModel(model)

    w.show()
    sys.exit(app.exec_())

```

srgblnch:
Then I will need a hint. TaurusBaseComponent.getDisplayValue(...) seems to only manage the fragment thing, to then call the TaurusBaseComponent.displayValue(...) that the TaurusValueLineEdit.setValue(...) was calling before. 

Using the modification I've made in the main of the tauruslineedit.py to allow just this testing, I call:
`$ python lib/taurus/qt/qtgui/input/tauruslineedit.py sys/tg_test/1/short_scalar sys/tg_test/1/short_scalar#rvalue.magnitude sys/tg_test/1/short_scalar#wvalue.magnitude`

And yes, when the patched code is called without a fragment the value shown is the rvalue. This may play with the fragmentName argument? Sorry but the help in the methods is not helping me to understand them.

srgblnch:
Perhaps the TaurusBaseWritableWidget.setModel(...) (that is not implemented and is using the one in the superclass TaurusBaseComponent) should modify the defaultFragmentName to change on the writable things from 'rvalue' to 'wvalue'?

srgblnch:
Would this third commit be good enough to prevent the behaviour of showing the rvalue when no fragment is shown?

By the way, the bugfix doesn't interfere the superclass, as we have spoken, in order avoid an API impact.

lib/taurus/qt/qtgui/input/tauruslineedit.py
```diff
@@ -201,6 +201,29 @@ def _stepBy(self, v):
         value = self.getValue()
         self.setValue(value + Quantity(v, value.units))
 
+    def getDisplayValue(self, cache=True, fragmentName=None):
```
cpascual: This is not critical , but... most taurus code avoids using super because of some issues with diamond inheritance schemes, so I would prefer the following implementation:

```python
def getDisplayValue(self, cache=True, fragmentName=None):
    if fragmentName is None and self.modelFragmentName is None:
        fragmentName = 'wvalue'
    return TaurusBaseWidget.getDisplayValue(self, cache=cache, fragmentName=fragmentName)
```

lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -2061,6 +2061,30 @@ def getOperationCallbacks(self):
...
+        :return: (str) a string representation of the model value.
+        """
+        # @fixme: Even the widgets inheriting from this class interacts with
+        #         writable models (then the fragmentName by default that one
+        #         expects is 'wvalue' and not 'rvalue') it has been considered
+        #         that perhaps this could be considered an API modification
+        #         instead of a bugfix.
+        #         The bugfix impact has been bounded only within the
+        #         TaurusValueLineEdit.
+        return super(TaurusBaseWritableWidget,
```
cpascual: I would avoid using super (see my other comment)

lib/taurus/qt/qtgui/input/tauruslineedit.py
srgblnch: Then this TaurusValueLineEdit.getDisplayValue() may call the newer TaurusBaseWritableWidget.getDisplayValue() instead of the supersuperclass. Even it's an empty call right now, if this is propagated up one would realise better the scale.

cpascual: 100% right. It should be TaurusBaseWritableWidget.getDisplayValue

My mistake

cpascual:
Merging, with two minor changes.

Thanks @srgblnch !

PR #672, opened 2018-01-23
Add tangoState property to tango device objects

cpascual:
Add TangoDevice.tangoState. It is a property that returns the value of
the state attribute of the device.

This allows, e.g. to set the bgRole of a TaurusLabel to display the color of the *tango* state instead of the taurus state:

```python

if __name__ == '__main__':
    import sys
    app = TaurusApplication()
    w = TaurusLabel()
    # Show the status attr in the foreground
    w.setModel('sys/tg_test/1/status')
    # show the *tango* state color as background
    w.setBgRole('parentObj.tangoState')
    w.show()
    sys.exit(app.exec_())
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ develop branch) won't be reflected in this file.
 - Travis-built docs (not yet replacing the RTD ones) (#572)
 - TaurusLed now supports non-boolean attributes (#617)
 - Support for arbitrary bgRole in labels (#629)
+- `TangoDevice.tangoState` property (#672)
 - `--import-ascii` option in `taurusplot` launcher (#632)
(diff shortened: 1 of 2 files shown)


cpascual:
Ready for integration. @taurus-org/integrators , anyone?

cmft:
Hi @cpascual ,
I review the code and it is OK but you can get the same behavior with:
`w.setBgRole('parentObj.stateObj.rvalue')`

IMO this new property is not necessary

cpascual:
@cmft :
> you can get the same behavior with: `w.setBgRole('parentObj.stateObj.rvalue')`
IMO this new property is not necessary

You are absolutely right! 
I was looking for this solution Yesterday and for some reason I couldn't get it...

So I am rejecting this PR since it is unnecessary.

PR #673, opened 2018-01-23
(Fix #667) Problems editing text of TaurusValueLineEdit

### LINKED ISSUE #667: TaurusValueLineEdit does not allow to delete value
The way to reproduce is simple:
1- open a taurus form with a Read/write attribute with units
2- Put the cursor at the end of the write widget and try to delete the units.

It is not possible to delete the units.

3. Selecting all the value+units and pressing "SUPR" has no effect

Moreover

4- Pressing twice "DEL" the current value is restored

Taurus version: 4.3.0-alpha

Manolo

cpascual:
The changes done in #650 introduce a problem when editing text in
TaurusValueLineEdit widgets (see #667). 

Furthermore, it seems that the problem described in https://github.com/taurus-org/taurus/pull/650#issue-283282985 cannot be reproduced as described (see https://github.com/taurus-org/taurus/pull/650#issuecomment-359861563).

Therefore, revert the main change of #650 in order to fix #667.

### DIFF
--- a/lib/taurus/qt/qtgui/input/tauruslineedit.py
+++ b/lib/taurus/qt/qtgui/input/tauruslineedit.py
@@ -126,12 +126,6 @@ class TaurusValueLineEdit(Qt.QLineEdit, TaurusBaseWritableWidget):
         TaurusBaseWritableWidget.updateStyle(self)
 
         value = self.getValue()
-        if value is None and self.hasPendingOperations():
-            try:
-                value = self.getModelValueObj().wvalue
-                self.setValue(value)
-            except:
-                value = None
 
         if value is None or not self.isTextValid():
             # invalid value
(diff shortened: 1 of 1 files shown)


sergirubio:
The bug described in #650 is easily reproducible using the ctlyrtech gui; the line edits in the second column appear empty instead of showing the w_value.

cpascual:
Can you please provide a way of reproducing using standard widget (if possible TaurusValueLineEdit) and TangoTest DS?

sergirubio:
The problem with the widget is that getValue() is not able to differentiate an invalid text from a text that has never been set (it returns None for both cases).

My proposal is to modify the behaviour of getValue() in TaurusValueLineEdit:

Option A: When text is not valid, raising an exception in getValue() instead of returning None. 

Option B: Storing every setValue() as ._value ; this would allow to check for self._value = None.

sergirubio:
Hi, I have pushed the option B code in these branch:  
https://github.com/sergirubio/taurus/tree/tauruslineedit_bug_bis

sergirubio:
https://github.com/taurus-org/taurus/pull/674

cpascual:
The changes proposed in this PR have been incorporated into #674, so I am closing this PR.

PR #674, opened 2018-01-25
Fix several issues in TaurusValueLineEdit

### LINKED ISSUE #649: WritableWidgets do not update in case of errors
The TaurusBaseWidget is writing a none value ( --- ) in case of receiving error events.

But the TaurusBaseWritableWidget is ignoring them, it processes Change and Periodic events but ignores the errors.

In a TaurusLineEdit, it may induce to error as user may keep trying to write values on a device that is not running at all. The widget should show that the device is not there anymore.

### LINKED ISSUE #667: TaurusValueLineEdit does not allow to delete value

sergirubio:
On behalf of @mbroseta and @cpascual 

solves problems with nones and units after (re)connection with event-less devices

#649: line edit ignoring errors from device
#650: value not loaded without events
#667, #673: side effects of the previous fix
CS-14995: line edit and value label ignoring format on device reconnection

All the previous errors where reproduced using:

a Tango device with a RW attribute, with "deg" units and %6.2e format

and this code as client:

```python
import taurus

taurus.Factory('tango').set_tango_subscribe_enabled(False)

m = 'sys/tg_test/1/float_scalar'
a = taurus.Attribute(m)

app = TaurusApplication()

w = TaurusValueLineEdit()
w.setFormat(tangoFormatter)
w.setModel(m)
w.show()

app.exec_()
```

Fix #649 , 
Fix #650 , 
Fix #667, 
Fix #673 
Fix CS-14995

### DIFF
--- a/lib/taurus/qt/qtgui/input/tauruslineedit.py
+++ b/lib/taurus/qt/qtgui/input/tauruslineedit.py
@@ -127,6 +127,7 @@ class TaurusValueLineEdit(Qt.QLineEdit, TaurusBaseWritableWidget):
 
         if self._value is None and self.hasPendingOperations():
             try:
+                self.getModelObj().getAttributeInfoEx(cache = False)
                 value = self.getModelValueObj().wvalue
                 self.info('Overwriting wvalue=None with %s' % (value))
                 self.setValue(value)
--- a/lib/taurus/qt/qtgui/input/tauruslineedit.py
+++ b/lib/taurus/qt/qtgui/input/tauruslineedit.py
@@ -105,6 +105,19 @@ class TaurusValueLineEdit(Qt.QLineEdit, TaurusBaseWritableWidget):
         self.emitValueChanged()
 
     def handleEvent(self, evt_src, evt_type, evt_value):
+
+        # handle the case in which the line edit is not yet initialized
+        if self._last_value is None:
+            try:
+                self.getModelObj().read(cache=True)
+                frag_name = self.modelFragmentName or 'wvalue'
+                value = self.getModelFragmentObj(fragmentName=frag_name)
+                self.debug('Overwriting wvalue=None with %s' % (value))
+                self.setValue(value)
+                self.setEnabled(value is not None)
+            except Exception as e:
+                self.debug('Failed attempt to initialize value: %r', e)
+
         self.setEnabled(evt_type != TaurusEventType.Error)
         if evt_type in (TaurusEventType.Change, TaurusEventType.Periodic):
             self._updateValidator(evt_value)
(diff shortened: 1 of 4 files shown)


lib/taurus/qt/qtgui/input/tauruslineedit.py
```diff
@@ -124,15 +128,21 @@ def isTextValid(self):
     def updateStyle(self):
         TaurusBaseWritableWidget.updateStyle(self)
 
-        value = self.getValue()
-        if value is None and self.hasPendingOperations():
+        if self._last_value is None and self.hasPendingOperations():
+            # This check must be done at updateStyle() because pending
+            # operations list is not updated yet at handleEvent() stage
             try:
+                self.getModelObj().getAttributeInfoEx(cache = False)
```
cpascual: `getAttributeInfoEx()` is tango-centric. It should not be used here.

lib/taurus/qt/qtgui/input/tauruslineedit.py
```diff
@@ -124,15 +128,21 @@ def isTextValid(self):
...
                 value = self.getModelValueObj().wvalue
+                self.info('Overwriting wvalue=None with %s' % (value))
                 self.setValue(value)
```
cpascual: I think that updateStyle is not the right place for setting the value because:

- it may be called very often
- it is likely to produce bugs such as those seen in #650

cpascual: IMHO, we really should try to think of a way to do this kind of stuff in handleEvent

cpascual:
I merged the changes of #673 into this PR, so I am closing #673

Note that as it is now, this PR is still WIP. A few TODO's need to be solved before merging

cpascual:
All pending issues have been addressed in the latest commits. So I remove the WIP tag.

@sergirubio and @mbroseta, can you re-test if all the reported issues are now fixed?

reszelaz:
@sergirubio and @mbroseta did you have time to test it?

mbroseta:
Yes, I have already confirmed hte request and it's working well.

Many thanks.

reszelaz:
I have rerun the tests on travis (we have observed false positives - more details in #683). The tests still passes. So I'm merging this PR..

sergirubio:
dzi&#281;ki!

PR #678, opened 2018-01-29
Fix bug in setting serialization mode

cpascual:
TaurusModel's logic for setting the serialization mode is wrong due to
a tabulation problem. This leads to TaurusDevices always being
initialized with serialization mode 'Serial', ignoring the mode set
in the Manager (by default, 'Concurrent'). Fix the tabulation issue and
restore the intended logic.

Warning: be aware that because the Concurrent mode was in practice not
set until now (in spite of being the intended situation), some code may
not have been properly tested under Concurrent serialization mode and
therefore potentially bugs which have been hidden may now suface.

The problem can be seen with the following code (it prints `Concurrent=False` unless this patch is applied):

```python
import taurus
m = taurus.Manager()
m.setSerializationMode(taurus.core.TaurusSerializationMode.Concurrent)
a = taurus.Attribute('sys/tg_test/1/double_scalar')
print "Concurrent=", a._serialization_mode==taurus.core.TaurusSerializationMode.Concurrent
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ develop branch) won't be reflected in this file.
 - Make spyder dependency optional (#556)
 
 ### Fixed
+- TaurusModel ignoring the serialization mode (#678)
 - Wrong "missing units" warnings for non-numerical attributes (#580)
 - Taurus3 backwards compatibility issues (#496, #550)
 - False positives in taurus.check_dependencies (#612)
--- a/lib/taurus/core/taurusmodel.py
+++ b/lib/taurus/core/taurusmodel.py
@@ -60,7 +60,7 @@ class TaurusModel(Logger):
             s_obj = parent
             if s_obj is None:
                 s_obj = self.factory()
-                serializationMode = s_obj.getSerializationMode()
+            serializationMode = s_obj.getSerializationMode()
         self._serialization_mode = serializationMode
 
         self._parentObj = parent


cpascual:
The practical effects of changing from serial to concurrent mode can be seen after applying this PR with the following code (note: first you need to configure the `sys/tg_test/1/double_scalar` attribute in jive to send change events ):

```python
import time
import taurus

busy =False

def f(*a):
    global busy
    print "\n\nf:", time.time(),
    if busy:
        print "--discard--\n"
        return
    busy = True
    print "--sleeping--\n"
    time.sleep(5)
    busy = False

# Uncomment the following 2 lines to set the mode to serial (once this PR is applied)
# m = taurus.Manager()
# m.setSerializationMode(taurus.core.TaurusSerializationMode.Serial)

a = taurus.Attribute('sys/tg_test/1/double_scalar')
a.addListener(f)

for i in xrange(13):
    time.sleep(1)
    print '.',
```

reszelaz:
These two messages refer to the Taurus serialization mode:
https://sourceforge.net/p/tauruslib/taurus-devel/message/35256010/
https://sourceforge.net/p/tauruslib/taurus-devel/message/35293948/

cpascual:
Ok, now that the jan18 release is out, we want to integrate this one (we didn't want to do it just before the release because this was potentially disrupting).

Also, we decided to be conservative and change opt for making the **Tango scheme** serial **by default** (so that things should not break after merging this PR, even if they are not preapared for concurrent mode). Note that one can always set the serialization mode for individual models.

Issue #648, opened 2017-12-18
taurusLabel.setModelIndex() not working


CorF4:
Hello,

in my tango device server I am using some spectrum attributes, some are boolean, some are double or strings. I can display the boolean spectrum attributes  via taurusLed and setting the modelIndex to the corresponding index of the attribute. This works fine. However, I also want to display the string and double spectrum attributes. Using the same method does not work here. Instead of the single entry corresponding to modelIndex the whole array is displayed (if the taurusLabel is large enough). I am wondering if there is a bug with taurusLabel and spectrum attributes (I tested the boolean version with taurusLabel and it didn't work, but I am not 100% sure it is down to taurusLabel or non-booleans). I wanted to ask here first if you can help with that before I need to transform all my spectrum attributes to scalar attributes.

Thanks a lot!

cpascual:
Hi, I reproduced the behaviour that you describe with this simple snippet:

```python
from taurus.qt.qtgui.application import TaurusApplication
from taurus.qt.qtgui.display import TaurusLabel

if __name__ == '__main__':
    import sys
    app = TaurusApplication()

    w = TaurusLabel()

    model = 'eval:10*arange(3)'
    w.setModel(model)
    w.setModelIndex(2)

    w.show()
    sys.exit(app.exec_()
```

The output is a label whose text is `[0 10 20]` instead of the expected `20`.

I think this is a bug in TaurusLabel (it used to work at some time, but it is not a commonly used feature and we did not realize the regression). 

We should fix it and also add unit tests to prevent regressions (but I cannot guarantee to devote time to it before the next release in January). If someone else wants to help...

CorF4:
Hi, thanks for the confirmation!
Indeed, I see that it works in taurus-3.6.0.

PhilLAL:
Hi,
reading the doc (http://www.taurus-scada.org/en/latest/devel/api/taurus/qt/qtgui/display/_TaurusLabel.html#taurus.qt.qtgui.display.TaurusLabel.modelIndex) I wonder what is the expected behaviour for modelIndex?

cpascual:
It is for array attrs. Showing only one item instead of a whole array.

cpascual:
It is for array attrs. Showing only one item instead of a whole array.

PhilLAL:
Ok, thank you for the explanation. 
I suggest completing the doc like this for the newcomer: «This property holds the index inside the model value that should be displayed when the model value is an array attribute.»

[...]

PR #681, opened 2018-02-12
Proposing fix for issue #648 (taurusLabel.setModelIndex() not working).

mkomel:
Fix is really simple, instead of calling getDisplayValue on the label it should be done on "self", so that it ends up in TaurusScalarAttributeControllerHelper.getDisplayValue, where the modelIndex handling is implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ develop branch) won't be reflected in this file.
 - Missing icons in buttons (#583, #598)
 - Exception in TaurusCommandForm (#608)
 - Launchers not showing output on MS Windows (#644)
-- Various issues with input widgets (#623, #661, #650, #669)
+- Various issues with input widgets (#623, #661, #650, #669, #651, #663)
 - Regressions in:
   - TaurusTrend (#618)
   - TaurusGrid (#609)
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -640,6 +640,14 @@ class TaurusBaseComponent(TaurusListener, BaseConfigurableClass):
             fragmentName = self.modelFragmentName
         return self.modelObj.getFragmentObj(fragmentName)
 
+    def getModelIndexValue(self):
+        """
+        Called inside getDisplayValue to use with spectrum attributes. By default not used, but some widget might want to support this feature.
+
+        Override when needed.
+        """
+        return None
+
     def getFormatedToolTip(self, cache=True):
         """Returns a string with contents to be displayed in a tooltip.
 
@@ -817,6 +825,11 @@ class TaurusBaseComponent(TaurusListener, BaseConfigurableClass):
             v = self.getModelFragmentObj(fragmentName=fragmentName)
         except:
             return self.getNoneValue()
+
+        idx = self.getModelIndexValue()
+        if v is not None and idx is not None and len(idx):
+            for i in idx:
+                v = v[i]
 
         return self.displayValue(v)
 
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -642,7 +642,9 @@ class TaurusBaseComponent(TaurusListener, BaseConfigurableClass):
 
     def getModelIndexValue(self):
         """
-        Called inside getDisplayValue to use with spectrum attributes. By default not used, but some widget might want to support this feature.
+        Called inside getDisplayValue to use with spectrum attributes.
+        By default not used, but some widget might want to support this
+        feature.
 
         Override when needed.
         """
@@ -827,7 +829,7 @@ class TaurusBaseComponent(TaurusListener, BaseConfigurableClass):
             return self.getNoneValue()
 
         idx = self.getModelIndexValue()
-        if v is not None and idx is not None and len(idx):
+        if v is not None and idx:
             for i in idx:
                 v = v[i]
 
--- a/lib/taurus/qt/qtgui/base/tauruscontroller.py
+++ b/lib/taurus/qt/qtgui/base/tauruscontroller.py
@@ -257,7 +257,7 @@ class TaurusConfigurationControllerHelper(object):
             try:
                 no_val = getattr(model, "no_" + param)  # TODO: Tango-centric
                 if val.lower() == no_val.lower():
-                    val = widget.getNoneValue()
+                    return widget.getNoneValue()
             except:
                 pass
         except AttributeError:
@@ -278,13 +278,13 @@ class TaurusConfigurationControllerHelper(object):
                     val = val.replace('<dev_name>', dev.getNormalName() or
                                       '---')
             else:
-                val = widget.getNoneValue()
+                return widget.getNoneValue()
         except:
             widget.debug("Invalid configuration parameter '%s'" % param)
-            val = widget.getNoneValue()
+            return widget.getNoneValue()
         if val is None:
-            val = widget.getNoneValue()
-        return val
+            return widget.getNoneValue()
+        return widget.displayValue(val)
 
 
 StyleSheetTemplate = """border-style: outset; border-width: 2px; border-color: {0}; {1}"""
--- a/lib/taurus/qt/qtgui/base/tauruscontroller.py
+++ b/lib/taurus/qt/qtgui/base/tauruscontroller.py
@@ -243,7 +243,7 @@ class TaurusConfigurationControllerHelper(object):
     @property
     def configParam(self):
         if self._configParam is None:
-            self._configParam = self.widget().getModelFragmentName() or ''
+            self._configParam = self.widget().modelFragmentName or ''
         return self._configParam
 
     def getDisplayValue(self, write=False):
--- a/lib/taurus/qt/qtgui/base/tauruscontroller.py
+++ b/lib/taurus/qt/qtgui/base/tauruscontroller.py
@@ -175,25 +175,25 @@ class TaurusAttributeControllerHelper(object):
 
 class TaurusScalarAttributeControllerHelper(TaurusAttributeControllerHelper):
 
-    def getDisplayValue(self, write=False):
+    def getDisplayValue(self, write=False, fragmentName=None):
         valueObj = self.valueObj()
         widget = self.widget()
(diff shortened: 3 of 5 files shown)


mkomel:
Fixes #648

cpascual:
Ho, thanks a lot for the proposed fix.

Unfortunately, this fix breaks some functionality (the support for fragments):

For example, this code raises an AttributeError:
```python

if __name__ == '__main__':
    import sys
    app = TaurusApplication()
    w = TaurusLabel()
    model = 'bl97/tangotest/1/float_scalar#label'
    w.setModel(model)
    w.show()
    sys.exit(app.exec_())
```

I am looking into it to try to find a solution

cpascual:
Note. This PR was based on an old-ish checkout (previous to the integration of #564).
I am merging the latest develop into this PR (and we will see that the unit tests are no longer passing)

cpascual:
I am surprised that the Travis tests appear green. This worries me because locally they fail:

```
 :...taurus(mkomel-develop)$  python -m unittest -v taurus.qt.qtgui.display.test.test_tauruslabel` 

(...)
Ran 35 tests in 99.831s
FAILED (failures=3, errors=8)
```

Looking at the [line 4813 of the logs](https://travis-ci.org/taurus-org/taurus/jobs/341365584#L4813), I see: 
```
$ travis_wait docker exec -t taurus-test /bin/bash -c "TAURUS_STARTER_WAIT=5 taurustestsuite -e 'taurus\.core\.util\.test\.test_timer'"
Still running (1 of 20): docker exec -t taurus-test /bin/bash -c TAURUS_STARTER_WAIT=5 taurustestsuite -e 'taurus\.core\.util\.test\.test_timer'
The command docker exec -t taurus-test /bin/bash -c TAURUS_STARTER_WAIT=5 taurustestsuite -e 'taurus\.core\.util\.test\.test_timer' exited with 0.
Log:
/home/travis/.travis/job_stages: line 169:  4725 Terminated              travis_jigger $! $timeout $cmd
The command "travis_wait docker exec -t taurus-test /bin/bash -c "TAURUS_STARTER_WAIT=5 taurustestsuite -e 'taurus\.core\.util\.test\.test_timer'"" exited with 0.
```

And the bit about `/home/travis/.travis/job_stages: line 169:  4725 Terminated              travis_jigger $! $timeout $cmd` worries me (maybe we are silently ignoring test failures in travis now?)

mkomel:
By looking a bit further, the controller that the tauruslabel uses in case when fragments are used seems to be of type TaurusConfigurationControllerHelper. This controller expects getModelFragmentName to exist on the widget, but TaurusLabel does not have it.

Since there is no widget in taurus that actually has getModelFragmentName function, I think it must have been a mistake during some past refactoring to not align this with what the widgets have (e.g. either use modelFragmentName field on the widgets, either add function getModelFragmentName on the widgets). 

I have made a commit with the first solution (changing a call .getModelFragmentName() to .modelFragmentName)

mkomel:
In order for fragments like att_name#rvalue.units  to work, additional change had to be made (previous commit).

Note that some tauruslabel tests still fail, however I don't really understand why those tests are written as they are:
as one example: 
```
@insertTest(helper_name='checkFormat',
                    model='eval:Q("5m")#rvalue.units',
                    formatter=baseFormatter2,
                    expected="Unit")
```
Why would the above test expect "Unit"? If I use this model in code, it shows 'm' as one would expect, and not 'Unit'.

mkomel:
Ok, this seems to do it (commit #b456fb1). Formatting of the value created in TaurusConfigurationControllerHelper is now delegated to the widget (actually very similar to what and how it is done in TaurusBase.getDisplayValue, which is where the code goes prior to my proposed fix).
The tests now pass.

It seems to me that there is some duplication or ambiguity in general in code around TaurusLabel - should it use stuff from the TaurusXXXControllerHelper(s) or stuff from TaurusBase? There are things that are implemented properly in one and not the other, and vice versa. For example, modelIndex handling was done in TaurusXXXControllerHelper, while fragments support and formatting in TaurusBase.  I guess the code should be merged somehow (e.g. to have a common implementation of features for widgets in one place).

cpascual:
> The tests now pass.

Some tests are still failing on my machine.

```
$ python -m unittest -v taurus.qt.qtgui.display.test.test_tauruslabel.TaurusLabelTest2
(...)
======================================================================
FAIL: test_text_2 (taurus.qt.qtgui.display.test.test_tauruslabel.TaurusLabelTest2)
Testing text(expected='double_scalar', model='tango:TangoSchemeTest/unittest/temp-1/double_scalar', fgRole='label')
  File "/nfs/home/cpascual/src/taurus/lib/taurus/test/base.py", line 113, in newTest
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/display/test/test_tauruslabel.py", line 157, in text
AssertionError: wrong text for "tango:TangoSchemeTest/unittest/temp-1/double_scalar":
 expected: double_scalar
 got: 1.23 mm

======================================================================
FAIL: test_text_3 (taurus.qt.qtgui.display.test.test_tauruslabel.TaurusLabelTest2)
Testing text(expected='double_scalar', model='tango:TangoSchemeTest/unittest/temp-1/double_scalar#rvalue', fgRole='label')
  File "/nfs/home/cpascual/src/taurus/lib/taurus/test/base.py", line 113, in newTest
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/display/test/test_tauruslabel.py", line 157, in text
AssertionError: wrong text for "tango:TangoSchemeTest/unittest/temp-1/double_scalar#rvalue":
 expected: double_scalar
 got: 1.23 mm

======================================================================
FAIL: test_text_7 (taurus.qt.qtgui.display.test.test_tauruslabel.TaurusLabelTest2)
Testing text(expected='ATTR_VALID', model='tango:TangoSchemeTest/unittest/temp-1/double_scalar', maxdepr=1, fgRole='quality')
  File "/nfs/home/cpascual/src/taurus/lib/taurus/test/base.py", line 113, in newTest
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/display/test/test_tauruslabel.py", line 157, in text
AssertionError: wrong text for "tango:TangoSchemeTest/unittest/temp-1/double_scalar":
 expected: ATTR_VALID
 got: 1.23 mm

======================================================================
FAIL: test_text_9 (taurus.qt.qtgui.display.test.test_tauruslabel.TaurusLabelTest2)
Testing text(expected='0.00 mm', model='tango:TangoSchemeTest/unittest/temp-1/double_scalar', maxdepr=1, fgRole='w_value')
  File "/nfs/home/cpascual/src/taurus/lib/taurus/test/base.py", line 113, in newTest
  File "/nfs/home/cpascual/src/taurus/lib/taurus/qt/qtgui/display/test/test_tauruslabel.py", line 157, in text
AssertionError: wrong text for "tango:TangoSchemeTest/unittest/temp-1/double_scalar":
 expected: 0.00 mm
 got: 1.23 mm

Ran 14 tests in 32.052s

FAILED (failures=4)

```

cpascual:
> It seems to me that there is some duplication or ambiguity in general in code around TaurusLabel - should it use stuff from the TaurusXXXControllerHelper(s) or stuff from TaurusBase? There are things that are implemented properly in one and not the other, and vice versa. For example, modelIndex handling was done in TaurusXXXControllerHelper, while fragments support and formatting in TaurusBase. I guess the code should be merged somehow (e.g. to have a common implementation of features for widgets in one place).

You are 100% right. The `*ControllerHelper` stuff is there for historical reasons (they where an attempt at generalization that was never further developed) and IMHO we should eventually get rid of them by refactoring and simplifying the display widgets.

mkomel:
I overlooked those TestSchemeTest tests.. I have done some additional changes, now the tests truly should pass.
I made changes to tauruscontroller.py (adding fragmentName to functions), then also had to tweak the overriden function signature in tauruslcd.py, in order to not break that widget.

Overall, it now works (as far as I can see), but it feels just a bit hacked up. I'm wondering whether it would make more sense to just implement modelindex handling in taurusbase  (maybe something to note down as a comment somewhere in code?) Especially since you are saying that you plan to get rid of the *ControllerHelper classes at some point..
But anyway, since the current solution seem to work I would stick with it for now.

cpascual:
I just brought the latest changes from develop that fix the travis unittest issues detected in this PR before (see #683).

mkomel:
I have now reverted the changes and instead done the modelindex handling in taurusbase.py.
This seems like a cleaner solution, the tests now also pass.

cpascual:
I think everything is ok now (I just pushed a couple of minor changes to your last commit). 
Merging into develop.
Thanks @mkomel for the contribution (it wasn't an easy part of Taurus you got involved with!)  ;)

CorF4:
Hello all,
coming back to this: I noticed that there is a problem with tango attributes in getDisplayValue() if the attribute quality is ATTR_INVALID (that is, we use push_change_event with attribute quality as argument). It seems that in this case ` v` is` []`, which is not `None` but the `idx` will then be out of range, so we need a protection against that: I propose to use simply (in taurusbase.py, lines 831 following):

```
        idx = self.getModelIndexValue()
        if v is not None and idx:
            if len(v):
                for i in idx:
                    v = v[i]
            else:
                return self.getNoneValue()
```

cpascual:
I think that it is better (clearer and more efficient) to protect with a try-except.

CorF4:
I agree with the try-except block, but I think we cannot simply write v = v[idx]: I see that in the case we have a string spectrum attribute we have tuples here and if I use your proposal the fields constantly show the none-value. Only idea I have is to use `v = v[idx[0]]` instead of the for-loop.

cpascual:
> in the case we have a string spectrum attribute we have tuples here and if I use your proposal the fields constantly show the none-value. 

You are right. my proposed code only makes sense if both v and idx are numpy arrays. 

> Only idea I have is to use v = v[idx[0]] instead of the for-loop.

I think that to be closer to the behaviour of the previous implementation (loop)  the correct thing would be:
`v = v[idx[-1]] `

> I could do a pull-request: do I have permission to create a branch?

In Taurus we use the fork-and-pull workflow (see https://github.com/taurus-org/taurus/blob/develop/CONTRIBUTING.md ) so you do not need permission in the official repo to do a PR.

CorF4:
Ok, thanks. Pull request is submitted  here #729.

PR #684, opened 2018-02-16
Add mechanism to discover taurus.qt.qtgui plugins

cmft:
Setuptools provides special support for plugins (entry_points).
Plugins can register themselves for discovery.

Add mechanism to discover and load all of the registered entry
points for the taurus.qt.qtgui module by using
"pkg_resources.iter_entry_points" method, and expose them as part of
the module.

### DIFF
--- a/lib/taurus/qt/qtgui/__init__.py
+++ b/lib/taurus/qt/qtgui/__init__.py
@@ -38,6 +38,7 @@ import sys
 import glob
 import pkg_resources
 from taurus import tauruscustomsettings as __S
+from taurus import debug as __debug
 
 icon_dir = os.path.join(os.path.dirname(os.path.abspath(__icon.__file__)))
 # TODO: get .path file glob pattern from tauruscustomsettings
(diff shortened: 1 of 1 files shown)


cpascual:
I changed the code to "hide" it as much as possible in order to avoid unaware people relying on an experimental API

cpascual:
Note that this can be considered a proof-of-concept for a possible implementation of [TEP13](http://www.taurus-scada.org/tep?TEP13.md)

cmft:
This is an example of setup for plugins

```
from setuptools import setup, find_packages

setup(
    name='taurus_pyqtgraph',
    packages=find_packages(),
    entry_points={'taurus.qt.qtgui': 'tpg = taurus_pyqtgraph'},
)
```

cpascual:
This solution works fpr doing `from taurus.qt.qtgui import foo`, but it does not work with `import  taurus.qt.qtgui.foo`

cpascual:
This is ready for merging, IMHO but since I made mods myself, I would prefer @cmft or some @taurus-org/integrators to give an ok before merging to develop

cmft:
It seems ok for me

PR #687, opened 2018-02-19
Fixing TaurusLed handling of spectrum attributes.

mkomel:
This proposal fixes TaurusLed handling of spectrum boolean attributes, which was broken by the following commit (from November 2017):
[eb17e2ed33195bf08aa57673fe0f40016ffdeb87](https://github.com/taurus-org/taurus/commit/eb17e2ed33195bf08aa57673fe0f40016ffdeb87)

See also comment under the above commit for detailed explanation of the problem.

### DIFF
--- a/lib/taurus/qt/qtgui/display/taurusled.py
+++ b/lib/taurus/qt/qtgui/display/taurusled.py
@@ -77,26 +77,19 @@ class _TaurusLedController(object):
         value = None
         if fgRole == 'rvalue':
             value = obj.rvalue
-            if type(value) == int:
-                value = bool(value)
         elif fgRole == 'wvalue':
             value = obj.wvalue
-            if type(value) == int:
-                value = bool(value)
         elif fgRole == 'quality':
             return obj.quality
 
         # handle 1D and 2D values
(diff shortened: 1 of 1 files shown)


cpascual:
Hi, thanks for the proposed fix, but I see some issues with the implementation:
- using `type(foo) == sometype` is discouraged (using isinstance instead is slightly better, but even that should be avoided in favour of [ducktyping](https://en.wikipedia.org/wiki/Duck_typing) whenever possible)
- the conversion should not only be done for `int`. In fact, IMHO, a cast to bool should be done in all cases, and the support for non-scalar attributes should be based on applying the modelIndex slicing before the cast

mkomel:
Agreed, commited alternative solution, taking into account your comments.

mkomel:
This fix has not been merged yet.. 

Since the problem still exists in the latest release, could you please consider merging it?

cpascual:
Hi @mkomel:  sorry for the delay. I get into re-reviewing it right now...

cpascual:
All ok for me. Merging. Thanks for contributing, and sorry again for the delay in reviewing.

Cheers!

PR #690, opened 2018-02-19
Allow to remove log handlers from Logger

reszelaz:
Add method `removeLogHandler` to provide API for removing log handlers.

Whenever accepted I will create another PR to support-3.x


reszelaz:
I'm closing this PR cause the right one #691 is opened to the release-Jan18 branch.

PR #694, opened 2018-02-20
Allow case sensitivity for epics object

napppoli:
Currently, case sensitivity for epics object is set False and "_lowerIfInsensitive" function on taurusplot.py modifies PV name to lower case.

```
if scheme == "scan" or not taurus.Factory(scheme).caseSensitive:
    models.append(str(m).lower())
 ```

As a result, model name does not match with those PVs whose name contain upper cases, and TaurusPlot and TaurusTrend fail retrieving their data.

This proposal sets caseSensitive true  and enables epics PVs containing upper case in their name to be displayed on TaurusPlot and TaurusTrend.

### DIFF
--- a/lib/taurus/core/epics/epicsfactory.py
+++ b/lib/taurus/core/epics/epicsfactory.py
@@ -57,7 +57,7 @@ class EpicsFactory(Singleton, TaurusFactory, Logger):
     schemes = ("ca", "epics",)
     DEFAULT_DEVICE = 'ca:'
     DEFAULT_AUTHORITY = 'ca://'
-    caseSensitive = False
+    caseSensitive = True
     elementTypesMap = {TaurusElementType.Authority: EpicsAuthority,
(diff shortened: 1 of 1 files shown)


cpascual:
I didn't realize that case sensitiveness was disabled in the taurus epics scheme. Merging. Thanks for the fix!

@nappoli : just out of curiosity: since there are not many users of epics in Taurus, it would be great if you could share about your experience with taurus+epics (how and where do you use it, what are your plans for it, etc.)

napppoli:
Hi @cpascual , I am currently working at [Cosylab d.d.](https://www.cosylab.com/) , a company in Slovenia that provides control system developement service using mainly EPICS. 
I've been testing out Taurus to find out what functions are implemented/ not implemented so far, for the future possibility of using Taurus as an alternative EPICS GUI in some of our future project.
I hope I can be of any help in improving the EPICS scheme of Taurus!

Issue #668, opened 2018-01-17
Model Chooser does not respect Case sensitivity of models when editing the list (double-click)


guifrecuni:
Hi, I found an issue when trying to change models using Model Chooser (both in TaurusForm and TaurusTrend applications).

You can reproduce it by:
1) Run TaurusForm with models that contain upper- and lower-case letters like:
taurusform 'eval:"AbCd"' sys/database/2/StoredProcedureRelease

2) right-click to open context menu, and select the "Modify Contents" option to open the Model Chooser

3) doubleclick on any model in the list to modify it

cpascual:
Confirmed

cpascual:
The problem is caused by the following line in `TaurusForm.chooseModels()`:

```python
models = [m.lower() for m in self.getModel()]
```

This is done in order to ensure a unique ID in the case of case-insensitive models.
A better solution could be implemented if #779 is implemented

cpascual:
PR #780 has been done to fix this issue (independently of #779)

PR #697, opened 2018-02-21
Convert model name list to CaselessList

napppoli:
This commit will convert model name list to CaselessList in case it is not one.

Current code does not cause any problem if you launch TaurusTrend only once. However, if you create TaurusGUI and create a TaurusTrend panel, it fails retrieving models whose name contains upper case from second launch because of the del_sets.

This proposal will prevent model list that contains upper cases in their name but shouldn't be removed from being removed by del_sets.

### DIFF
--- a/lib/taurus/qt/qtgui/plot/taurustrend.py
+++ b/lib/taurus/qt/qtgui/plot/taurustrend.py
@@ -1157,6 +1157,8 @@ class TaurusTrend(TaurusPlot):
         try:
             # For it to work properly, 'names' must be a CaselessList, just as
             # self.trendSets is a CaselessDict
+            if not isinstance(names, CaselessList):
+                names = CaselessList(names)
             del_sets = [name for name in self.trendSets.keys()
                         if name not in names]
 


cpascual:
Hi @napppoli , sorry for the late reply (we are deep in work with the Jan18 release).

On a quick glance, I think that your proposed solution may not be adequate when dealing with case sensitive schemes (e.g. the `eval` scheme ). But please give us more time to review this properly once we get the release done.

cpascual:
Hi, I am finally looking into this. Here is a quick way to reproduce the bug:

1. Launch :
```
taurustrend 'SYS/TG_TEST/1/SHORT_SCALAR' \
                    'sys/tg_test/1/float_scalar' \
                    'eval:@builtins.*/ord("A")' \
                    'eval:@builtins.*/ord("a")' \
                    'eval:@builtins.*/ord("B")' \
                    'eval:@builtins.*/ord("c")' \
                    -r200
```

2. open the input data selection dialog (CTRL+N)
3. click on apply in the dialog 
4. click on apply in the dialog (again)

Without this patch, the  `ord("B")` curve disappears in step 3 and reappears in step 4
With this patch, the above issue is solved

But, in any case, the support for case-sensitive curves is not good (with or without the patch) because:

- a) the `ord("a")`curve is not shown: it is wrongly treated as a duplicate of `ord("A")`
- b) the case of `SYS/TG_TEST/1/SHORT_SCALAR` is not moaintained for display
- c) #668 


Nevertheless, this PR does not seem to break anything, so I am merging it.
Thanks @napppoli for the contribution!

PR #712, opened 2018-02-28
Fix infinite recursion error when debugging TaurusValueSpinBox

### LINKED ISSUE #708: Infinite recursion error when debugging TaurusValueSpinBox
While debugging #707 we hit a strange recursion error if we try to put a print in `TaurusValueSpinBox.validate()`

The problem can be reproduced with the following snippet:

```python
from taurus.qt.qtgui.input import TaurusValueSpinBox
import taurus

class A(TaurusValueSpinBox):
    def validate(self, input, pos):
        val = self.lineEdit().validator()
        taurus.debug('asd')
        if val is None:
            return Qt.QAbstractSpinBox.validate(self, input, pos)
        return val.validate(input, pos)

app = TaurusApplication()
w = A()
model = 'sys/tg_test/1/double_scalar'
w.setModel(model)
```

When running it, I get:

```
  File "/home/cpascual/.PyCharmCE2017.2/config/scratches/scratch_68.py", line 11, in validate
    taurus.debug('asd')
  File "/home/cpascual/src/taurus/lib/taurus/core/util/log.py", line 1019, in debug
    return __getrootlogger().debug(msg, *args, **kw)
  File "/home/cpascual/src/taurus/lib/taurus/core/util/log.py", line 1004, in __getrootlogger
    return Logger.getLogger("TaurusRootLogger")
  File "/home/cpascual/src/taurus/lib/taurus/core/util/log.py", line 656, in getLogger
    cls.initRoot()
  File "/home/cpascual/src/taurus/lib/taurus/core/util/log.py", line 526, in initRoot
    return cls._getLogger()
  File "/home/cpascual/src/taurus/lib/taurus/core/util/log.py", line 652, in _getLogger
    logging.setLoggerClass(orig_logger_class)
  File "/usr/lib/python2.7/logging/__init__.py", line 996, in setLoggerClass
    if klass != Logger:
RuntimeError: maximum recursion depth exceeded in cmp
(...) (REPEATED SEVERAL TIMES)
```

By using git bisect, the problem seems to be starting with 69c39a6 and, indeed, if we comment out the call to `setValue` in the first `if` block of `TaurusValueLineEdit.handleEvent()` (which was introduced with 69c39a6) this problem is gone.

cmft:
The handleEvent of TaurusValueLineEdit try to validate a value
(setValue) before updating properly the validator and this causes
the infinite recursion.

Call updateValidator before evalueate the event value.

Fix #708


cpascual:
Hi tnaks, but I tried this and it does not fix the issue for me (tested with the snippet from https://github.com/taurus-org/taurus/issues/708#issue-300594435)

cpascual:
... but it does fix a sys except hook error .

Nevertheless, the change proposed here is already done in 72ecf52 , so I am closing this one as dup.

Issue #711, opened 2018-02-28
Win7 Taurus4.3: check_dependencies does not detect spyder >3 in some occasions


mrosanes:
Testing in Win7 with Taurus4.3, with pythonxy installed, and with a subsequent installation of spyder done with:
pip install --upgrade --no-deps spyder,

The version of the installed spyder without dependencies being, 3.2.7 (the import is done correctly from ipython)

The installation is not detected correctly by taurus.check_dependencies()
[ ] spyder>=3

--

Note: if installing with dependencies (again spyder 3.2.7):
pip install --upgrade spyder
Then the installation is detected correctly by taurus.check_dependencies()
[*] spyder>=3

tiagocoutinho:
I don't think this is a bug. Doing `pip install --no-deps spyder` is not the proper way to install spyder. I would say that after this command spyder should not be considered to be installed.

This behaviour seems to be consistent between linux and windows

cpascual:
Agreed. I close as rejected.

PR #721, opened 2018-03-13
Remove unneeded URI adaptation in taurusgui

cmft:
In createInstrumentsFromPool method taurus complete the URI
with the scheme. After sardana FQDN changes (#711) it is not
longer necessary.


cpascual:
Hi, thanks for the PR, but instead of removing the adaptation, I'll make it conditional to maintain support for sardana <2.4
Also, I'll do it directly as a hotfix for Taurus. So I am closing this.

PR #723, opened 2018-03-15
Remove 3rd party code in taurus.external

### LINKED ISSUE #85: taurus.external modules may not give proper warnings
This is a minor issue affecting the imports from taurus.external of the enum, ordereddict and pint modules.

The mechanism for importing the taurus.external.enum module is implemented as follows in the `__init__.py` (for example in the case of pint):

```~~
from __future__ import absolute_import

try:
    from pint import *
except ImportError:
    # pint from local import
    import os
    import sys
    import warnings
    warnings.warn("pint not available. Using local pint", ImportWarning)
    sys.path.append(os.path.dirname(__file__))
    from pint import *
    del warnings
    del sys
    del os
```~~

By reading this code, one would expect to get a warning every time the local copy of pint needs to be imported, but it is not the case in the following situation:
a) if the user has the '.' dir added to their PYTHONPATH (a default configuration that most people have)
b) the name of the locally provided module matches that of the system library (this is the case for the three modules mentioned).

For example: with the current implementation, the pint module will be searched in the PYTHONPATH and, even if not installed elsewere, for most people it will be found in the '.' dir without raising an exception (and therefore the local copy will be used without warning).

See a more lengthy discussion here:
http://pythonquirks.blogspot.com.es/2010/07/absolutely-relative-import.html

I propose to reimplement along the following lines:

  * __rename__ taurus/external/pint/pint --> taurus/external/pint/_pint
  * then implement the `__init__.py` as follows:

    ```~~
    from __future__ import absolute_import
    
    try:
        from pint import *
    except ImportError:
        # pint from local import
        from warnings import warn as __warn
        __warn("pint not available. Using local pint", ImportWarning)
        from ._pint import *
    ```~~


### PR DESCRIPTION
**Pull request by cpascual**
Remove embedded 3rd-party code in taurus.external.argparse  taurus.external.pint,  taurus.external.enum and use dependencies instead.

Deprecate the corresponding wrapper modules and avoid using them internally in taurus.

This closes #85

### CODE REVIEW

**Comment by cpascual** [comment]
Note: the appveyor build is failing. I think that the issue is with the new install-require of pint>=0.8:

[From appveyor's log](https://ci.appveyor.com/project/taurusorg/taurus/build/job/f2j0kg2dxp1s5su9?fullLog=true#L9854):
```
Collecting pint>=0.8 (from taurus)
  Could not find a version that satisfies the requirement pint>=0.8 (from taurus) (from versions: )
No matching distribution found for pint>=0.8 (from taurus)
```

mrosanes:
Reviewed this pull request:
- Code
- taurusform with an eval Quantity
- Enumeration
- Different imports (OrderedDict, argparse, pint...) 
- Check deprecation warnings (proposed change from taurus.core.util to taurus.core.units)


From my point of view, the PR is Ok and it can be integrated.

PR #724, opened 2018-03-20
[doc] Move docs to github pages

### LINKED ISSUE #408: [doc] Get rid of mocks for docs

cpascual:
This PR does the last things pending for migrating the doc build and hosting from readthedocs to github pages.

See https://github.com/taurus-org/taurus/issues/408#issuecomment-329782603

Basically it eliminates the mocks and it adds a link to other versions of the docs.

Note: with this migration we will lose the downloadable pdf docs, but IMHO it can be added later on if someone is interested on it.

Closes #408


cpascual:
@taurus-org/integrators can you please review this?

Apart from reviewing the changes themselves, basically, 2 things need to be checked:

- the docs are served fine from github pages: https://taurus-org.github.io/taurus-doc
- the dynamically-generated list to docs generated for branches other than develop works fine: https://taurus-org.github.io/taurus-doc/v-fix-408/other_versions (this needs javascript allowed in the browser)

It all seems to work well for me... once I get confirmation that it works for others too, I'll merge and ask for the taurus-scada domain to be pointed to https://taurus-org.github.io/taurus-doc/ instead of readthedocs (and we will finally be able to get rid of mocks and RTD other RTD support burden)

mrosanes:
From my point of view, it can be merged,
Nice feature to have the docs in github pages!

cpascual:
Great, thanks for the reviews @teresanunez and @mrosanes .

I am merging.

cpascual:
Checked. It all seems ok now.  Both taurus-scada.org and www.taurus-scada.org point to GH pages

I also deactivated the RTD integration (RTD won't be receiving events from this repo anymore.

PR #725, opened 2018-03-20
Remove build badges from doc page

cpascual:
The main doc page contains badge images showing the license, the
CI build status, etc. Lintian complains about them when packaging
the docs due to privacy concerns:
https://lintian.debian.org/tags/privacy-breach-generic.html
Remove the badges from the html docs.

Note: the same badges are still present in the README file, which is 
rendered in the project page. But this is ok since it is not packaged as html

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -44,35 +44,6 @@ interfaces.
 :Release: |release|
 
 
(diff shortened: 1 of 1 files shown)


cpascual:
@taurus-org/integrators this is a trivial change IMHO. Can I go ahead and merg it?

mrosanes:
It seems ok to me as well

PR #726, opened 2018-03-20
Make taurus-3.x branch deploy docs github pages

cpascual:
Configure the support-3.x branch to deploy docs to github.com/taurus-org/taurus-doc so that it is served in https://taurus-org.github.io/taurus-doc/v-support-3.x/

(Prior to merging, they can be seen in https://taurus-org.github.io/taurus-doc/v-deploydoc3x/ )

This is complementary to #724


mrosanes:
Reviewed, it seems Ok to me

PR #728, opened 2018-03-23
Try to populate TaurusCommandForm even if device not ready

### LINKED ISSUE #727: TaurusDevicePanel doesn't show commands when device has FAULT state
When I want to open a `TaurusDevicePanel` and the device is in FAULT state, the commands don't show up. 
taurusdevicepanel-fault

cpascual:
This is a fix for #727. 

TaurusCommandForm conservatively clears its contents if the device is not in "Ready"
state. Relax this limitation and try to populate it anyway, falling back to clear in case of 
problems querying the commands.

I've tested this solution  with the TangoTest device but I am calling for help for other people 
testing it with other devices (not just TangoTest) to check if the proposed solution is safe enough.

@taurus-org/integrators (and anyone else, of course), can you have a look at it?

Fixes #727

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusform.py
+++ b/lib/taurus/qt/qtgui/panel/taurusform.py
@@ -636,13 +636,19 @@ class TaurusCommandsForm(TaurusWidget):
         Inserts command buttons and parameter widgets in the layout, according to
         the commands from the model
         '''
-        #self.debug('In TaurusCommandsForm._updateCommandWidgets())')
+
         dev = self.getModelObj()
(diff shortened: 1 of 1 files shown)


teresanunez:
Hi, I have tried it with another devices and I do not see any problem.

cpascual:
Thanks, @teresanunez . Do not be shy and merge it yourself if you think it is ok  ;)

MikeFalowski:
Everything is fine now. Thanks for fast reaction, guys!

PR #729, opened 2018-03-23
Bug fix in taurusbase.py concerning model indices (extension of pull request #681)

CorF4:
Added try-except block to prevent crashes if spectrum model returned is [] (happens if ATTR_INVALID in tango).

### DIFF
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -830,9 +830,11 @@ class TaurusBaseComponent(TaurusListener, BaseConfigurableClass):
 
         idx = self.getModelIndexValue()
         if v is not None and idx:
-            for i in idx:
-                v = v[i]
-
+            try:
+                v = v[idx[-1]]
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


cpascual:
The proposed  changes look ok to me (see discussion in https://github.com/taurus-org/taurus/pull/681#issuecomment-375280345 and next comment). Still pending to actually test it.

CorF4:
Isn't there one problem after all removing the foor-loop: in case of 2d attributes the idx object will contain two indices and to extract the correct value, we need the loop or ask for v[idx[0]][idx[1]]...is that correct?

cpascual:
> Isn't there one problem after all removing the foor-loop: in case of 2d attributes the idx object will contain two indices and to extract the correct value, we need the loop or ask for v[idx[0]][idx[1]]...is that correct?

Yes you are totally right. I've made a change to use the for-loop (but still protecting it with the try-except block).

Note: in any case, the TEP15 (#764) is about to be merged in the following days, and then I recommend to use the model-based slicing instead of the modelIndex property (I would even consider deprecating the modelIndex feature)

PR #738, opened 2018-04-13
Delegate the fireEvents to TaurusManager

cmft:
Use the TaurusManager PoolThread to manage the fireEvents
respecting the "serialization mode" of the scheme.

This PR delegates the control of the event queue to Taurus instead of Tango
when the serialization mode is Serial.

### DIFF
--- a/lib/taurus/core/evaluation/evalattribute.py
+++ b/lib/taurus/core/evaluation/evalattribute.py
@@ -484,7 +484,7 @@ class EvaluationAttribute(TaurusAttribute):
         if len(self._listeners) > 1 and \
            (initial_subscription_state == SubscriptionState.Subscribed or
                 self.isPollingActive()):
-            Manager().addJob(self.__fireRegisterEvent, None, (listener,))
+            Manager().enqueueJob(self.__fireRegisterEvent, None, (listener,))
         return ret
 
     def removeListener(self, listener):
--- a/lib/taurus/core/evaluation/evalattribute.py
+++ b/lib/taurus/core/evaluation/evalattribute.py
@@ -484,7 +484,8 @@ class EvaluationAttribute(TaurusAttribute):
         if len(self._listeners) > 1 and \
            (initial_subscription_state == SubscriptionState.Subscribed or
                 self.isPollingActive()):
-            Manager().enqueueJob(self.__fireRegisterEvent, None, ((listener,),))
+            Manager().enqueueJob(self.__fireRegisterEvent,
+                                 job_args=((listener,),))
         return ret
 
     def removeListener(self, listener):
--- a/lib/taurus/core/evaluation/evalattribute.py
+++ b/lib/taurus/core/evaluation/evalattribute.py
@@ -484,7 +484,7 @@ class EvaluationAttribute(TaurusAttribute):
         if len(self._listeners) > 1 and \
            (initial_subscription_state == SubscriptionState.Subscribed or
                 self.isPollingActive()):
-            Manager().enqueueJob(self.__fireRegisterEvent, None, ((listener,)))
+            Manager().enqueueJob(self.__fireRegisterEvent, None, ((listener,),))
         return ret
 
     def removeListener(self, listener):
--- a/lib/taurus/core/evaluation/evalattribute.py
+++ b/lib/taurus/core/evaluation/evalattribute.py
@@ -484,7 +484,7 @@ class EvaluationAttribute(TaurusAttribute):
         if len(self._listeners) > 1 and \
            (initial_subscription_state == SubscriptionState.Subscribed or
                 self.isPollingActive()):
-            Manager().enqueueJob(self.__fireRegisterEvent, None, (listener,))
+            Manager().enqueueJob(self.__fireRegisterEvent, None, ((listener,)))
         return ret
 
     def removeListener(self, listener):
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -591,7 +591,7 @@ class TangoAttribute(TaurusAttribute):
                 self.__fireRegisterEvent((listener,))
             else:
                 Manager().enqueueJob(self.__fireRegisterEvent, None,
-                                     jobargs=((listener,)),
+                                     jobargs=((listener,),),
                                      serialization_mode=sm)
         return ret
 
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -583,9 +583,15 @@ class TangoAttribute(TaurusAttribute):
         if len(listeners) > 1 and\
                 (initial_subscription_state == SubscriptionState.Subscribed or
                      self.isPollingActive()):
-
-            Manager().addJob(self.__fireRegisterEvent, None, (listener,),
-                             taurus_serialization_mode=self._serialization_mode)
+            sm = self._serialization_mode
+            if sm == TaurusSerializationMode.Serial:
+                self.deprecated('TaurusSerializationMode.Serial mode',
+                                alt="TaurusSerializationMode.SerialSync",
+                                rel='4.3.2')
+                self.__fireRegisterEvent((listener,))
+            else:
+                Manager().addJob(self.__fireRegisterEvent, None, (listener,),
+                                 taurus_serialization_mode=sm)
         return ret
 
     def removeListener(self, listener):
@@ -783,9 +789,16 @@ class TangoAttribute(TaurusAttribute):
                 return
             manager = Manager()
             listeners = tuple(self._listeners)
-            manager.addJob(self.fireEvent, None, etype, evalue,
-                           listeners=listeners,
-                           taurus_serialization_mode=self._serialization_mode)
+            sm = self._serialization_mode
+            if sm == TaurusSerializationMode.Serial:
+                self.deprecated('TaurusSerializationMode.Serial mode',
+                                alt="TaurusSerializationMode.SerialSync",
+                                rel='4.3.2')
+                self.fireEvent(etype, evalue, listeners=listeners)
+            else:
+                manager.addJob(self.fireEvent, None, etype, evalue,
+                               listeners=listeners,
+                               taurus_serialization_mode=sm)
 
     def _pushAttrEvent(self, event):
         """Handler of (non-configuration) events from the PyTango layer.
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -584,9 +584,9 @@ class TangoAttribute(TaurusAttribute):
                 (initial_subscription_state == SubscriptionState.Subscribed or
                      self.isPollingActive()):
             sm = self._serialization_mode
-            if sm == TaurusSerializationMode.Serial:
-                self.deprecated('TaurusSerializationMode.Serial mode',
-                                alt="TaurusSerializationMode.SerialSync",
+            if sm == TaurusSerializationMode.TangoSerial:
+                self.deprecated(dep='TaurusSerializationMode.TangoSerial mode',
+                                alt="TaurusSerializationMode.Serial",
                                 rel='4.3.2')
                 self.__fireRegisterEvent((listener,))
             else:
@@ -790,9 +790,9 @@ class TangoAttribute(TaurusAttribute):
             manager = Manager()
             listeners = tuple(self._listeners)
             sm = self._serialization_mode
-            if sm == TaurusSerializationMode.Serial:
-                self.deprecated('TaurusSerializationMode.Serial mode',
-                                alt="TaurusSerializationMode.SerialSync",
+            if sm == TaurusSerializationMode.TangoSerial:
+                self.deprecated(dep='TaurusSerializationMode.TangoSerial mode',
+                                alt="TaurusSerializationMode.Serial",
                                 rel='4.3.2')
                 self.fireEvent(etype, evalue, listeners=listeners)
             else:
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -581,8 +581,9 @@ class TangoAttribute(TaurusAttribute):
 
         # if initial_subscription_state == SubscriptionState.Subscribed:
         if (len(listeners) > 1
-            and (initial_subscription_state == SubscriptionState.Subscribed or
-                     self.isPollingActive())):
+            and (initial_subscription_state == SubscriptionState.Subscribed 
+                 or self.isPollingActive())
+           ):
             sm = self._serialization_mode
             if sm == TaurusSerializationMode.TangoSerial:
                 self.deprecated(dep='TaurusSerializationMode.TangoSerial mode',
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -580,9 +580,9 @@ class TangoAttribute(TaurusAttribute):
             self._subscribeEvents()
 
         # if initial_subscription_state == SubscriptionState.Subscribed:
-        if len(listeners) > 1 and\
-                (initial_subscription_state == SubscriptionState.Subscribed or
-                     self.isPollingActive()):
+        if (len(listeners) > 1
+            and (initial_subscription_state == SubscriptionState.Subscribed or
+                     self.isPollingActive())):
             sm = self._serialization_mode
             if sm == TaurusSerializationMode.TangoSerial:
                 self.deprecated(dep='TaurusSerializationMode.TangoSerial mode',
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -586,12 +586,12 @@ class TangoAttribute(TaurusAttribute):
             sm = self._serialization_mode
             if sm == TaurusSerializationMode.TangoSerial:
                 self.deprecated(dep='TaurusSerializationMode.TangoSerial mode',
-                                alt="TaurusSerializationMode.Serial",
+                                alt='TaurusSerializationMode.Serial',
                                 rel='4.3.2')
                 self.__fireRegisterEvent((listener,))
             else:
-                Manager().addJob(self.__fireRegisterEvent, None, (listener,),
-                                 taurus_serialization_mode=sm)
+                Manager().enqueueJob(self.__fireRegisterEvent, None,
+                                     (listener,), serialization_mode=sm)
         return ret
 
     def removeListener(self, listener):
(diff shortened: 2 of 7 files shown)


lib/taurus/core/tango/tangoattribute.py
```diff
@@ -580,12 +580,18 @@ def addListener(self, listener):
             self._subscribeEvents()
 
         # if initial_subscription_state == SubscriptionState.Subscribed:
-        if len(listeners) > 1 and (initial_subscription_state == SubscriptionState.Subscribed or self.isPollingActive()):
-            sm = self.getSerializationMode()
-            if sm == TaurusSerializationMode.Concurrent:
-                Manager().addJob(self.__fireRegisterEvent, None, (listener,))
-            else:
+        if len(listeners) > 1 and\
```
cpascual: minor PEP8 comment: avoid using `\`. Suggestion: 

```python
if (len(listeners) > 1 
         and (initial_subscription_state == SubscriptionState.Subscribed
                or self.isPollingActive() 
                ) 
    ):
```

lib/taurus/core/taurusmanager.py
```diff
@@ -82,13 +82,16 @@ def reInit(self):
...
+
+        self._thread_pool = ThreadPool(name="TaurusTP",
+                                       parent=self,
+                                       Psize=5,
+                                       Qsize=1000)
+
+        self._sthread_pool = ThreadPool(name="TaurusTSP",
+                                       parent=self,
+                                       Psize=1,
+                                       Qsize=float("inf")) #TODO check value
```
cpascual: Can you put a float there?. Also, PEP8: add extra space before and after `#`

lib/taurus/core/taurusmanager.py
```diff
@@ -120,15 +125,28 @@ def addJob(self, job, callback=None, *args, **kw):
         :param args: (list) list of arguments passed to the job
         :param kw: (dict) keyword arguments passed to the job
         """
-        if self._serialization_mode == TaurusSerializationMode.Concurrent:
+        if kw.has_key("taurus_serialization_mode"):
```
cpascual: Do we need to support adding the serialization mode kwarg here? (the previous implementation did not require it...)
If we do, at least we should document it in the method docstring, because ` taurus_serialization_mode` is not passed to the job (which is what the docstring says)

lib/taurus/core/taurusmanager.py
```diff
@@ -120,15 +125,28 @@ def addJob(self, job, callback=None, *args, **kw):
...
                 self.info("Job cannot be processed.")
                 self.debug(
-                    "The requested job cannot be processed. Make sure this manager is initialized")
+                    "The requested job cannot be processed. " +
+                    "Make sure this manager is initialized")
                 return
             self._thread_pool.add(job, callback, *args, **kw)
         else:
-            job(*args, **kw)
+            if not hasattr(self, "_sthread_pool") or \
```
cpascual: PEP8. Avoid using `\`. In fact, since you do not need the check on hasattr, you do not need to break the line

lib/taurus/core/taurusmanager.py
```diff
@@ -120,15 +125,28 @@ def addJob(self, job, callback=None, *args, **kw):
...
         :param kw: (dict) keyword arguments passed to the job
         """
-        if self._serialization_mode == TaurusSerializationMode.Concurrent:
+        if kw.has_key("taurus_serialization_mode"):
+            serialization_mode = kw.pop("taurus_serialization_mode")
+        else:
+            serialization_mode = self._serialization_mode
+
+        if serialization_mode == TaurusSerializationMode.Concurrent:
             if not hasattr(self, "_thread_pool") or self._thread_pool is None:
```
cpascual: the `hasattr` check is not needed anymore since you are now always initializing the pool.

lib/taurus/core/taurusmanager.py
cmft: I fixed the TODO

lib/taurus/core/taurusmanager.py
cmft: Yes, we need. The previous implementation treat the serial mode at the scheme level. So,  the scheme overwrite the setting at the scheme level and not at manager level

> If we do, at least we should document it in the method docstring, because taurus_serialization_mode is not passed to the job (which is what the docstring says)

Agree. I will

lib/taurus/core/taurusmanager.py
cmft: We saw that, but we repeated the same protection that _thread_pool has.

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +111,49 @@ def cleanUp(self):
...
                 self.info("Job cannot be processed.")
                 self.debug(
-                    "The requested job cannot be processed. Make sure this manager is initialized")
+                    "The requested job cannot be processed. " +
+                    "Make sure this manager is initialized")
                 return
-            self._thread_pool.add(job, callback, *args, **kw)
+            self._thread_pool.add(job, callback, jobargs, jobkwargs)
```
cpascual: I think that it should be `self._thread_pool.add(job, callback, *jobargs, **jobkwargs)`

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +111,49 @@ def cleanUp(self):
...
         else:
-            job(*args, **kw)
+            if (not hasattr(self, "_sthread_pool")
+                or self._sthread_pool is None):
+                self.info("Job cannot be processed.")
+                self.debug(
+                    "The requested job cannot be processed. " +
+                    "Make sure this manager is initialized")
+                return
+            self._sthread_pool.add(job, callback, jobargs, jobkwargs)
```
cpascual: I think that it should be self._thread_pool.add(job, callback, *jobargs, **jobkwargs)

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +111,49 @@ def cleanUp(self):
...
             if not hasattr(self, "_thread_pool") or self._thread_pool is None:
                 self.info("Job cannot be processed.")
                 self.debug(
-                    "The requested job cannot be processed. Make sure this manager is initialized")
+                    "The requested job cannot be processed. " +
+                    "Make sure this manager is initialized")
                 return
-            self._thread_pool.add(job, callback, *args, **kw)
+            self._thread_pool.add(job, callback, jobargs, jobkwargs)
         else:
```
cpascual: It should be `elif serialization_mode == TaurusSerializationMode.Serial:`

lib/taurus/core/taurusmanager.py
cpascual: ... and then thre should be another elif for supporting "TangoSerial" and finally an `else` for raising a `RuntimeError("Unsupported serialization mode {}".format(serialization_mode))`

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +111,49 @@ def cleanUp(self):
...
-        if self._serialization_mode == TaurusSerializationMode.Concurrent:
+        if serialization_mode is None:
+            serialization_mode = self._serialization_mode
+
+        if serialization_mode == TaurusSerializationMode.Concurrent:
             if not hasattr(self, "_thread_pool") or self._thread_pool is None:
                 self.info("Job cannot be processed.")
                 self.debug(
-                    "The requested job cannot be processed. Make sure this manager is initialized")
+                    "The requested job cannot be processed. " +
```
cpascual: PEP 8 recommends putting the "+" operator on the next line

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -580,12 +580,18 @@ def addListener(self, listener):
             self._subscribeEvents()
 
         # if initial_subscription_state == SubscriptionState.Subscribed:
-        if len(listeners) > 1 and (initial_subscription_state == SubscriptionState.Subscribed or self.isPollingActive()):
-            sm = self.getSerializationMode()
-            if sm == TaurusSerializationMode.Concurrent:
-                Manager().addJob(self.__fireRegisterEvent, None, (listener,))
-            else:
+        if (len(listeners) > 1
+            and (initial_subscription_state == SubscriptionState.Subscribed or
```
cpascual: PEP 8 recommends putting the "or" operator on the next line

cpascual:
The testsuite is failing systematically.: @cmft can you check if this is also the case when run locally (i.e. not in travis?)

lib/taurus/core/taurusmanager.py
cmft: > I think that it should be self._thread_pool.add(job, callback, *jobargs, **jobkwargs)
This is wrong but IMO it should be `self._thread_pool.add(job, callback, args=jobargs, kwargs=jobkwargs)`

See:

```
def foo(a,b, *args, **kwargs):
    print "foo", args, kwargs


def newfoo(job, callback, jobargs=None, jobkwargs=None, myfoo=None):
    # foo(job, callback, *jobargs, **jobkwargs) # TypeError: foo() argument after ** must be a mapping, not NoneType
    foo(job, callback, args=jobargs, kwargs=jobkwargs)


newfoo(1,2)
newfoo(1,2, (3, 4, 5), {'1': 1})
```

cpascual: Please have a look at the signature of `taurus.core.util.threadpool.ThreadPool.add()`

It is: `add(self, job, callback=None, *args, **kw)`

This means that the args and kwargs are expected "expanded" ... and we do not want to change that API now.

lib/taurus/core/taurusmanager.py
```diff
@@ -121,7 +121,7 @@ def addJob(self, job, callback=None, *args, **kw):
         """ Deprecated. Wrapper of enqueueJob. See enqueueJob documentation.
         """
         self.deprecated(dep='addJob', alt='enqueueJob', rel='4.3.2')
-        self.enqueueJob(job, callback=callback, jobargs=args, jobkwargs=kw)
+        self.enqueueJob(job, callback, args, kw)
```
cpascual: I do not agree with this change. It is better to be explicit and use kwargs, so the previous version is preferable.

lib/taurus/core/evaluation/evalattribute.py
```diff
@@ -484,7 +484,7 @@ def addListener(self, listener):
         if len(self._listeners) > 1 and \
            (initial_subscription_state == SubscriptionState.Subscribed or
                 self.isPollingActive()):
-            Manager().enqueueJob(self.__fireRegisterEvent, None, (listener,))
+            Manager().enqueueJob(self.__fireRegisterEvent, None, ((listener,)))
```
cpascual: It would be better to use kwargs explicitly. 
Also, the extra parentheses added are superfluous (note that since there is no `,`, they do not denote a tuple but just a superfluous priority).
In summary, it should be:
`Manager().enqueueJob(self.__fireRegisterEvent, callback=None, jobargs=(listener,))`

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -591,7 +591,8 @@ def addListener(self, listener):
                 self.__fireRegisterEvent((listener,))
             else:
                 Manager().enqueueJob(self.__fireRegisterEvent, None,
-                                     (listener,), serialization_mode=sm)
+                                     jobargs=((listener,)),
```
cpascual: Also, the extra parentheses added are superfluous . See note above

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -796,7 +797,9 @@ def push_event(self, event):
                                 rel='4.3.2')
                 self.fireEvent(etype, evalue, listeners=listeners)
             else:
-                manager.enqueueJob(self.fireEvent, None, etype, evalue,
+                manager.enqueueJob(self.fireEvent, None,
```
cpascual: It would be better and more consistent to pass the callback arg as a kwarg (i.e., use `callback=None` instead of `None`)

lib/taurus/core/evaluation/evalattribute.py
```diff
@@ -484,7 +484,7 @@ def addListener(self, listener):
         if len(self._listeners) > 1 and \
            (initial_subscription_state == SubscriptionState.Subscribed or
                 self.isPollingActive()):
-            Manager().addJob(self.__fireRegisterEvent, None, (listener,))
+            Manager().enqueueJob(self.__fireRegisterEvent, None, ((listener,)))
```
cpascual: It would be better to use kwargs explicitly.
Also, the extra parentheses added are superfluous (note that since there is no ,, they do not denote a tuple but just a superfluous priority).
In summary, it should be:
Manager().enqueueJob(self.__fireRegisterEvent, callback=None, jobargs=(listener,))

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -580,12 +580,19 @@ def addListener(self, listener):
...
+                     self.isPollingActive())):
+            sm = self._serialization_mode
+            if sm == TaurusSerializationMode.TangoSerial:
+                self.deprecated(dep='TaurusSerializationMode.TangoSerial mode',
+                                alt='TaurusSerializationMode.Serial',
+                                rel='4.3.2')
                 self.__fireRegisterEvent((listener,))
+            else:
+                Manager().enqueueJob(self.__fireRegisterEvent, None,
+                                     jobargs=((listener,)),
```
cpascual: Same as before: the extra parentheses added are superfluous .

lib/taurus/core/tango/tangoattribute.py
```diff
@@ -782,13 +789,18 @@ def push_event(self, event):
...
-                               listeners=listeners)
-            else:
+            sm = self._serialization_mode
+            if sm == TaurusSerializationMode.TangoSerial:
+                self.deprecated(dep='TaurusSerializationMode.TangoSerial mode',
+                                alt="TaurusSerializationMode.Serial",
+                                rel='4.3.2')
                 self.fireEvent(etype, evalue, listeners=listeners)
+            else:
+                manager.enqueueJob(self.fireEvent, None,
```
cpascual: It would be better and more consistent to pass the callback arg as a kwarg (i.e., use callback=None instead of None)

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +112,59 @@ def cleanUp(self):
...
 
         self._state = ManagerState.CLEANED
 
     def addJob(self, job, callback=None, *args, **kw):
-        """Add a new job (callable) to the queue. The new job will be processed
-        by a separate thread
+        """ Deprecated. Wrapper of enqueueJob. See enqueueJob documentation.
+        """
+        self.deprecated(dep='addJob', alt='enqueueJob', rel='4.3.2')
+        self.enqueueJob(job, callback, args, kw)
```
cpascual: It is better to use kwargs:
`self.enqueueJob(job, callback=callback, jobargs=args, jobkwargs=kw)`

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +112,59 @@ def cleanUp(self):
...
 
+    def enqueueJob(self, job, callback=None, jobargs=None, jobkwargs=None,
+                   serialization_mode=None):
+        """ Enqueue a job (callable) to the queue. The new job will be
+        processed by a separate thread
         :param job: (callable) a callable object
         :param callback: (callable) called after the job has been processed
-        :param args: (list) list of arguments passed to the job
-        :param kw: (dict) keyword arguments passed to the job
+        :param jobargs: (list) list of arguments passed to the job
```
cpascual: Use "sequence" instead of "list" in the documentation (in fact the default value for jobarg is not a list but a tuple!)

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +112,59 @@ def cleanUp(self):
...
+        :param jobkwargs: (dict) keyword arguments passed to the job
+        :param serialization_mode:
+        (taurus.core.taurusbasetypes.TaurusSerializationMode) serialization
+        mode
         """
-        if self._serialization_mode == TaurusSerializationMode.Concurrent:
+        if jobargs is None:
```
cpascual: Why not putting `jobargs=()` in the method signature? It would be clearer and shorter

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +112,59 @@ def cleanUp(self):
...
 
     def addJob(self, job, callback=None, *args, **kw):
-        """Add a new job (callable) to the queue. The new job will be processed
-        by a separate thread
+        """ Deprecated. Wrapper of enqueueJob. See enqueueJob documentation.
+        """
+        self.deprecated(dep='addJob', alt='enqueueJob', rel='4.3.2')
+        self.enqueueJob(job, callback, args, kw)
 
+    def enqueueJob(self, job, callback=None, jobargs=None, jobkwargs=None,
```
cpascual: PEP8: change jobargs to job_args and jobkwargs to job_kwargs

Note that this implies changes everywhere else where it is used!

lib/taurus/core/taurusmanager.py
```diff
@@ -108,27 +112,59 @@ def cleanUp(self):
...
+                    + "Make sure this manager is initialized")
+                return
+            self._thread_pool.add(job, callback, *jobargs, **jobkwargs)
+        elif serialization_mode == TaurusSerializationMode.Serial:
+            if (not hasattr(self, "_sthread_pool")
+                    or self._sthread_pool is None):
+                self.info("Job cannot be processed.")
+                self.debug(
+                    "The requested job cannot be processed. "
+                    + "Make sure this manager is initialized")
```
cpascual: Should we not consider also the case `elif serialization_mode == TaurusSerializationMode.TangoSerial:` ?

lib/taurus/core/tauruspollingtimer.py
```diff
@@ -108,7 +108,7 @@ def addAttribute(self, attribute, auto_start=True):
             self.start()
         else:
             import taurus
-            taurus.Manager().addJob(attribute.poll, None)
+            taurus.Manager().enqueueJob(attribute.poll, None)
```
cpascual: better use callback=None

lib/taurus/core/evaluation/evalattribute.py
cmft: > the extra parentheses added are superfluous

This was a merge mistake we need to pass a tuple of tuple in that case.

lib/taurus/core/taurusmanager.py
cmft: IMH it is not an agnostic option, so I will not considered as an option for TaurusManager. It should only be used by the Tango scheme as back. comp. option (deprecated)

lib/taurus/core/tango/tangoattribute.py
cmft: In fact I removed it, since it is not necessary and it is the default value

lib/taurus/core/taurusmanager.py
cpascual: ok. good point

cpascual:
Note: the tests were passing, but got interrupted when I merged the commit. No big deal.

PR #747, opened 2018-05-08
TaurusValueSpinBox behaviour with keyboard arrows

srgblnch:
We realise the keyboard arrows in the spinbox widget were not acting like the button arrows in the widget. The behaviour in taurus3 was the same both ways using the singleStep.

There is also a functionality in LineEdit widgets to use a modifier of the step using Ctrl to multiply by 10 the step and Alt to divide by 10, while they are pressed.

This patch like to fix the behaviour of the keyboard arrows as well as to introduce the modifiers.


cpascual:
Hi @srgblnch, thanks for the PR, but we would prefer a different approach. I am implementing the preferred approach on #749 (BTW, it would be great if you could test if it meets your needs), so I am closing this one.

PR #748, opened 2018-05-08
Handle datetime strings without microseconds

cpascual:
The TaurusPlot.importAscii method raises a ValueError when importing
timestamp strings written without decimals in the seconds.

For example:
```
ValueError: time data '2018-04-09_00:08:49' does not match format '%Y-%m-%d_%H:%M:%S.%f'
```

This absence of decimals has been observed in some cases with files
exported by TaurusTrend.

Make the isodatestr2float converter function more robust by allowing
to handle both cases (with and without decimals)

This can be tested with the following file: 
[foo.txt](https://github.com/taurus-org/taurus/files/1984236/foo.txt)


cpascual:
@

PR #749, opened 2018-05-16
 Improve consistency in stepping for TaurusValueSpinbox

cpascual:
TaurusValueSpinBox composes a TaurusValueLineEdit and delegates many of its features to it. But the "singleStep" property is not delegated and this causes inconsistent handling of step increments (e.g. pressing the keyboard up/down arrows ignores the step size, while using the up/down arrows of the spinbox widget takes it into account). 

Also, the use of the CTRL and ALT key to modify the step size is not consistently supported in all the possible step triggers (keyboard arrows, spinbox arrows and wheel events). And worse, both modifiers conflict with other roles (the control is used also for "force-applying" and the ALT+mouse for dragging windows)

Fix all these issues by refactoring the stepping feature of TaurusValueLineEdit (an make TaurusSpinbox delegate to it). Note that this involves *removing* the role as multiplier for the CTRL and ALT keys.

Note: this is an alternative solution to that suggested by @srgblnch  in #747

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,11 +25,14 @@ develop branch) won't be reflected in this file.
 ### Removed
 - All 3rd party code from taurus.external (now using dependencies
   instead of embeded 3rd party code)
+- CTRL and ALT keys are no longer used to modify step size in
+  TaurusValueLineEdit and TaurusValueSpinbox (#749)
 
 ## [4.3.1] - 2018-03-14
 A hotfix release needed for sardana 2.4
 
-### Fixed 
+### Fixed
+- consistency issues in stepping support in spinboxes and line edits (#749)
 - duplicated "tango://" prefix in panels created from Pool for sardana>=2.4
 - avoid problems if channel dimension info is set to None by sardana (#722)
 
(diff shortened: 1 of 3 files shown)


cpascual:
This can be tested with the following snippets:

For spinbox:
```python
from taurus.qt.qtgui.input import TaurusValueSpinBox

if __name__ == '__main__':
    import sys
    app = TaurusApplication()
    w = TaurusValueSpinBox()
    w.setSingleStep(0.1)
    model = 'sys/tg_test/1/double_scalar'
    w.setModel(model)
    w.resize(200,30)
    w.show()
    sys.exit(app.exec_())
```
For line edit:
```python

if __name__ == '__main__':
    import sys
    app = TaurusApplication()
    w = TaurusValueLineEdit()
    w.setSingleStep(0.1)
    w.setEnableWheelEvent(True)
    model = 'sys/tg_test/1/double_scalar'
    w.setModel(model)
    w.resize(200,30)
    w.show()
    sys.exit(app.exec_())
```

teresanunez:
Hi, I have tested it and it works. I merge it.

PR #752, opened 2018-05-24
Fix wrong evaluation

### LINKED ISSUE #750: TangoAttribute getRange does not work
I set the ampli attribute negative limit to 0 using Jive.

```
In [1]: import taurus
In [2]: ampli = taurus.Attribute("sys/tg_test/1/ampli")
In [3]: ampli.getRange()
Out[3]: [<Quantity(-inf, 'dimensionless')>, <Quantity(inf, 'dimensionless')>]

In [4]: import PyTango
In [5]: ampli_tango = PyTango.AttributeProxy("sys/tg_test/1/ampli")
In [6]: conf = ampli_tango.get_config()
In [7]: conf.min_value
Out[7]: '0'
```

The same applies to getMinValue.
It works with taurus 3 but what is strange to me that it does not work in any of Taurus 4 releases.

cmft:
TangoAttribute configure its range, warning and alarm attributes
using an "or" evaluation. This evaluation fails when a "0" value is present
in the tango db attributes; Causing unexpected "inf" values instead
(that means not  defined in taurus).

Fix #750

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,8 @@ A hotfix release needed for sardana 2.4
 ### Fixed 
 - duplicated "tango://" prefix in panels created from Pool for sardana>=2.4
 - avoid problems if channel dimension info is set to None by sardana (#722)
-
+- unexpected "inf" values in tangoAttribute range, warning and alarm 
+  attributes (#750)
 
 ## [4.3.0] - 2018-03-01
 [Jan18 milestone](https://github.com/taurus-org/taurus/milestone/10)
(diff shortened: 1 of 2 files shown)


cpascual:
good catch. The fix looks ok to me.
Can you please also update the changelog?

PS: Just one precision for the record: "inf" does **not** mean "not defined" here (or elsewhere in taurus, AFAIK). In this case the logic is that if the Tango DB does not specify a limit for a numeric, then the less restrictive limit is set (this is `float("-inf")` for lower limits and `float("inf")` for upper limits)

cpascual:
Thanks, merging.

PS: I am ignoring the test failure in jessie. It looks to me as related to #261 (although I've forced a rebuild 3 times already and it "stubbornly persists" .  Anyway, maybe it is about time to stop doing CI on the oldstable debian (jessie)

PR #754, opened 2018-05-24
Force a repaint after reset the format

### LINKED ISSUE #753: Set Formatter  option does not work for "static" attributes
Set Formatter  context menu option does not work for attributes that not send events. Even-thought the formatter has changed the widget still showing the value in the old mode. If you force a repaint in the widget (e.g. taurusform) it will be automatic applied.

IMO a repaint should be done after a formatter changed.

cmft:
The "static" attributes (whitout events) do not change the
representation after a Fotmatter changed.

Fix it, calling  getDisplayValue() after reset the format.

Fix #753

### DIFF
--- a/lib/taurus/qt/qtgui/display/tauruslabel.py
+++ b/lib/taurus/qt/qtgui/display/tauruslabel.py
@@ -90,6 +90,8 @@ class TaurusLabelController(TaurusBaseController):
         elif fgRole.lower() in ('', 'none'):
             pass
         else:
+            label.modelFragmentName = fgRole
+            label.resetFormat()
             value = label.getDisplayValue(fragmentName=fgRole)
         self._text = text = label.prefixText + value + label.suffixText
 
--- a/lib/taurus/qt/qtgui/display/test/test_tauruslabel.py
+++ b/lib/taurus/qt/qtgui/display/test/test_tauruslabel.py
@@ -94,47 +94,47 @@ testOldFgroles = functools.partial(insertTest, helper_name='text', maxdepr=1,
                                    model='tango:' + DEV_NAME + '/double_scalar')
 
 
-@testOldFgroles(fgRole='value', expected='1.23 mm')
-@testOldFgroles(fgRole='w_value', expected='0.00 mm')
-@testOldFgroles(fgRole='state', expected='Ready')
+# @testOldFgroles(fgRole='value', expected='1.23 mm')
+# @testOldFgroles(fgRole='w_value', expected='0.00 mm')
+# @testOldFgroles(fgRole='state', expected='Ready')
 @testOldFgroles(fgRole='quality', expected='ATTR_VALID')
-@testOldFgroles(fgRole='none', expected='')
+# @testOldFgroles(fgRole='none', expected='')
 # ------------------------------------------------------------------------------
-@insertTest(helper_name='text',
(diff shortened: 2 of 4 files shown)


cpascual:
Hi, I pushed a change to docstrings to reflect the new behaviour. 
Apart from that, I think the fix is ok.
Can you please also update the CHANGELOG?

cpascual:
Oops! and also note that a has been broken with this change:

```
======================================================================
FAIL: test_text_7 (taurus.qt.qtgui.display.test.test_tauruslabel.TaurusLabelTest2)
Testing text(expected='ATTR_VALID', model='tango:TangoSchemeTest/unittest/temp-1/double_scalar', maxdepr=1, fgRole='quality')
  File "/usr/local/lib/python2.7/dist-packages/taurus-4.3.2a0-py2.7.egg/taurus/test/base.py", line 113, in newTest
  File "/usr/local/lib/python2.7/dist-packages/taurus-4.3.2a0-py2.7.egg/taurus/qt/qtgui/display/test/test_tauruslabel.py", line 157, in text
AssertionError: wrong text for "tango:TangoSchemeTest/unittest/temp-1/double_scalar":
 expected: ATTR_VALID
 got: 0
```

lib/taurus/qt/qtgui/display/test/test_tauruslabel.py
```diff
@@ -94,47 +94,47 @@ def test_relativemodelclass(self):
                                    model='tango:' + DEV_NAME + '/double_scalar')
 
 
-@testOldFgroles(fgRole='value', expected='1.23 mm')
-@testOldFgroles(fgRole='w_value', expected='0.00 mm')
-@testOldFgroles(fgRole='state', expected='Ready')
+# @testOldFgroles(fgRole='value', expected='1.23 mm')
```
cpascual: I guess you commented out these tests while debugging but you forgot to uncoment them back....

lib/taurus/qt/qtgui/display/tauruslabel.py
```diff
@@ -90,6 +90,8 @@ def _updateForeground(self, label):
         elif fgRole.lower() in ('', 'none'):
             pass
         else:
+            label.modelFragmentName = fgRole
```
cpascual: I do not like the idea of altering the label.modelFragmentName every time that you call _updateForeground on its controller (this is a side-efect and it is asking for future trouble, IMHO).
At the very least, I think that the original fragment name should be resored just after calling resetFormat.

...and even that would look quite ugly.

Isn't there any more elegant solution?

lib/taurus/qt/qtgui/display/test/test_tauruslabel.py
cmft: You are right, I forget to remove the "kk" commit.

lib/taurus/qt/qtgui/display/tauruslabel.py
cmft: > Isn't there any more elegant solution?

I did not find it, the other solution is at the taurusform level calling  crl.update  method after set a new Format

cmft:
I close this PR since the necessary workaround can introduce more troubles. For the moment the manual  resizing of the GUI solves the problem

PR #755, opened 2018-05-24
Fixed right-click segfault on TaurusMainWindow gui's

sergirubio:
As TaurusMainWindow is the last container in the chain
it must accept() events instead of ignore(). When ignoring
it tries to propagate the event to parent widgets, that in
this case is Null and generates segfaults

### DIFF
--- a/lib/taurus/qt/qtgui/container/taurusmainwindow.py
+++ b/lib/taurus/qt/qtgui/container/taurusmainwindow.py
@@ -807,8 +807,19 @@ class TaurusMainWindow(Qt.QMainWindow, TaurusBaseContainer):
             self.socketServer.close()
         Qt.QMainWindow.closeEvent(self, event)
         TaurusBaseContainer.closeEvent(self, event)
+        
+        # print "\n\n------ MAIN WINDOW CLOSED ------ \n\n"        
+        
+    def contextMenuEvent(self, event):
+        """Handle the popup menu event, as it is the last container in the chain
+        it must accept() events instead of ignore()
 
-        # print "\n\n------ MAIN WINDOW CLOSED ------ \n\n"
+        :param event: the popup menu event
+        """
+        if self.taurusMenu is not None:
+            self.taurusMenu.exec_(event.globalPos())
+        else:
+            event.accept()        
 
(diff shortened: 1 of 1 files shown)


sergirubio:
https://jira.cells.es/browse/CS-15681

cpascual:
Hi @sergirubio , thanks for the proposed fix. However I would like to better understand what is going on, because:

1.  I could not reproduce it with a simple TaurusMainWindow : 
`python -m taurus.qt.qtgui.container.taurusmainwindow`
2. For the moment we could only reproduce the problem with certain (complex) GUIs in ALBA's production environment.
3. If the problem was as described above, QMainWindow would be affected as well, and it would be quite a big bug in PyQt, which seems a bit unlikely... so I would like to discard other possible explanations first

So I propose that we work a bit internally at ALBA into reproducing it in a more general environment and then we re-evaluate this PR.

sergirubio:
I guess that QMainWindow is not affected because it does not reimplement the contextMenuEvent hook (or do not calls event.ignore() within it), TaurusMainWindows does it when inheriting from TaurusBaseComponent

cpascual:
> I guess that QMainWindow is not affected because it does not reimplement the contextMenuEvent hook (or do not calls event.ignore() within it), TaurusMainWindows does it when inheriting from TaurusBaseComponent

That's a good tip. I actually checked if TaurusMainWindow was reimplementing the contextMenuEvent, but I wrongly assumed that its base classes did not reimplement it.... we should try to reproduce the problem taking this into account

cmft:
Taking a look into the affected GUI the problem seems to be related with unconnected menubars.  TaurusMainWindows has a taurusMenu that has an action connected to the mouse right click. I you do no wish this menu you can disable it declaring this class attribute in your reimplementations of TaurusMainWindows.

```
_showTaurusMenu = False 
```

sergirubio:
I think that the main idea is not having to declare anything to avoid 
segfaults

cmft:
> I think that the main idea is not having to declare anything to avoid
> segfaults
I agree, sometime ago I proposed to have a simple TaurusMainWindows without not generic menubar. But if you try to reuse a widget and modify it you have to take care of the side effects.

PR #764, opened 2018-06-18
[TEP15] Implement TEP15 (support slices in fragments)

### LINKED ISSUE #751: Support referencing to attribute slices in the model name
Hello,
I noticed when using the TaurusJDrawSynopticsView class that there is nothing foreseen to handle spectrum attributes with a given index (idx). In Tango it works for a DevState type if one puts [idx] at the end of the attribute name. Could it be useful to support this in taurus as well? And for all kind of numeric spectrum types?
I don't know how much this is actually used, but for my case I certainly need to make this work.

cpascual:
Implement [TEP15](http://www.taurus-scada.org/tep/?TEP15.md) (support referencing slices in model fragments).

Added unit tests. To see it in action, you can run

```python

if __name__ == '__main__':
    import sys
    app = TaurusApplication()

    w = TaurusLabel()

    model = 'sys/tg_test/1/float_spectrum#rvalue[10:5:-2]'
    w.setModel(model)

    m = w.getModelObj()
    m.write(range(0,2560,10))
    print m.getFragmentObj('rvalue[10:5:-2]')

    w.resize(333,33)
    w.show()
    sys.exit(app.exec_())
```

Fixes #751


cpascual:
@taurus-org/integrators : can you please review this and cast a vote? 
If accepted we could mark the TEP15 as ACCEPTED.

teresanunez:
IMO it can be accepted

reszelaz:
Accepting this PR and TEP15.

PR #703, opened 2018-02-23
Support for python 3 (+python 2)

piertoni:
After applying futurize stage1 I need some help on the following:
- Update Travis CI script for Python3
- on lib/taurus/qt/qtgui/display/qled.py i changed the following line in order to pass the test:

```
#def __init__(self, parent=None, ledsize=LedSize.SMALL, ledcolor=LedColor.GREEN):
def __init__(self, parent=None, ledsize=24, ledcolor=LedColor.GREEN):
```
It's an enumeration problem as the enumeration is not starting from 0.
For what I have seen can be something related to Taurus Enumerations that must start from 0.
Surely will be straightforward for some of you that knows the codebase well.

PR #680, opened 2018-01-31
Python2-3 compatibility

piertoni:
Some minor python2-3 compatibility concerning print function and exceptions

PR #765, opened 2018-06-28
Expression 'x' in Enumeration fails on python3 with KeyError.

asenchenko:
To solve this problem, explicit __contains__ method is added.

### DIFF
--- a/lib/taurus/core/util/enumeration.py
+++ b/lib/taurus/core/util/enumeration.py
@@ -142,6 +142,12 @@ class Enumeration(object):
         self._uniqueId += 1
         return n
 
+    def __contains__(self, i):
+        if isinstance(i, (int, long)):
+            return i in self.reverseLookup
+        elif isinstance(i, (str, unicode)):
+            return i in self.lookup
+
     def __getitem__(self, i):
         if isinstance(i, (int, long)):
             return self.whatis(i)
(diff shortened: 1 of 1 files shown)


cpascual:
Hi @asenchenko , I am surprised by this PR since Taurus does not support python 3 yet (see #680 ), although some effort has been started (see #703)
So, I wonder... are you using it (or some part of it) under python3 ? 
If so, could you give some details?
Also, you probably could join the conversation in #703 and contribute there (help is needed!)

Apart from that, since your commit is about `taurus.core.util.enumeration.Enumeration`, I think you should also be aware of [SEP12](http://www.sardana-controls.org/sep/?SEP12.md). While this is not officially implemented (we haven't implemented the backwards compatibility layer), in practice we try to follow it for new developments (i.e., **we try to base all new enumerations on the `Enum` module rather than on `taurus.core.util.enumeration.Enumeration`**)

Having said all that, I've revised your proposed changes and they all seem ok to me, so IMHO we can merge it, but before, could you please provide some code snippet demonstrating the intended feature/bugfix that it provides?

asenchenko:
Hi @cpascual. I'm sorry for the confusion. I know that Taurus does not support python3.
I'm porting Taurus to python3 and this bug appeared during this work. 
It fixes problem with enums in #703.

cpascual:
> I'm porting Taurus to python3 and this bug appeared during this work. It fixes problem with enums in #703.

Ok, then I wonder if it wouldn't make more sense to make a PR against the branch of #703 ? (just to centralise the python3-related effort and avoid merge conflicts due to parallel implementations).

However, for that we need to know if @piertoni is still interested in this... I'll ask in #703

cpascual:
Hi, I merged these changes into the branch for #703, and therefore I am closing this PR to centralise the discussion on #703 and #680

PR #766, opened 2018-07-02
Document TaurusValueLineEdit

### LINKED ISSUE #679: (doc) document how TaurusValueLineEdit parses quantities
It looks like white spaces are interpreted as multiplication sign when editing the value.

To reproduce simply run the default sample code:
(e.g.) python ~/workspace/taurus/lib/taurus/qt/qtgui/input/tauruslineedit.py
and type "2 3" (withouth the quotes) and then enter. You will see that the write value of "sys/tg_test/1/double_scalar" has been change to 6 (2*3)

cpascual:
Add TaurusValueLineEdit class docstring. Fixes #679


reszelaz:
This is how it looks now:

image

PR #767, opened 2018-07-02
Fix 647: simplify API for setting Formatter on forms

### LINKED ISSUE #647: Improve API for setting the formatter in TaurusForms
We have observed a different behavior of the units value when using or a Taurusform or writing the code. When using the Taurusform the units follow the formatter, but when writing the application by the code (below), the units show its long format instead of the tangoFormatter value.

```
import sys
from taurus.qt.qtgui.base import TaurusBaseComponent
TaurusBaseComponent.FORMAT = tangoFormatter
app = TaurusApplication(sys.argv)
tf = TaurusForm()
tf.setModel("sys/tg_test/1/short_scalar_ro")
tf.show()
app.exec_()
```

Br
Manolo.

cpascual:
Fix #647 by implementing the features described in https://github.com/taurus-org/taurus/issues/647#issuecomment-352401115 .

Setting the format at form level now works as expected:

```python
import sys
app = TaurusApplication(sys.argv)
tf = TaurusForm()
tf.setFormat(tangoFormatter)
tf.setModel(["sys/tg_test/1/short_scalar_ro", "sys/tg_test/1/ampli"])
tf.show()
app.exec_()
```

and setting it at single value level works as well:
```python
import sys
app = TaurusApplication(sys.argv)
tf = TaurusForm()
tf.setModel(["sys/tg_test/1/short_scalar_ro", "sys/tg_test/1/ampli"])
tf[0].setFormat(tangoFormatter)
tf.show()
app.exec_()
```

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusform.py
+++ b/lib/taurus/qt/qtgui/panel/taurusform.py
@@ -380,6 +380,16 @@ class TaurusForm(TaurusWidget):
                     rw.setFormat(format)
         return format
 
+    def setFormat(self, format):
+        """
+        Reimplemented to call setFormat on the taurusvalues
+        """
+        TaurusWidget.setFormat(self, format)
+        for item in self.getItems():
+            rw = item.readWidget()
+            if hasattr(rw, 'setFormat'):
+                rw.setFormat(format)
+
     def setCompact(self, compact):
         self._compact = compact
         for item in self.getItems():
@@ -472,6 +482,11 @@ class TaurusForm(TaurusWidget):
                 widget.setModifiableByUser(self.isModifiableByUser())
             except:
                 pass
+            try:
+                widget.setFormat(self.getFormat())
+            except Exception:
+                self.debug('Cannot set format %s to child %s',
+                           self.getFormat(), model)
             widget.setObjectName("__item%i" % i)
             self.registerConfigDelegate(widget)
             self._children.append(widget)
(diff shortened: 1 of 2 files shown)


cpascual:
Note that now the following also works and does not make the units be shown in long form (as reported in #647:

```
taurusform --default-formatter=taurus.core.tango.util.tangoFormatter \
                   sys/tg_test/1/short_scalar_ro sys/tg_test/1/ampli
```

cmft:
The code seems ok for me, but testing with taurusform the context menu does not work with format strings.

cmft:
> The  code seems ok for me, but testing with taurusform the context menu does not work with format strings.

Sorry testing with different number I figure out that It only affect to the ReadWidget

lib/taurus/qt/qtgui/panel/taurusform.py
```diff
@@ -380,6 +380,16 @@ def onSetFormatter(self):
                     rw.setFormat(format)
         return format
 
+    def setFormat(self, format):
```
reszelaz: I think that format is a Python built-in. In this case I would suggest to use `format_`.

lib/taurus/qt/qtgui/panel/taurusvalue.py
```diff
@@ -442,6 +445,14 @@ def onSetFormatter(self):
         if hasattr(self._readWidget, 'onSetFormatter'):
             return self._readWidget.onSetFormatter()
 
+    def setFormat(self, format):
```
reszelaz: I think that format is a Python built-in. In this case I would suggest to use `format_`

reszelaz:
I see that in the related code base there are many more `format` variable names so I leave it as it is.

PR #770, opened 2018-07-12
Use simple name in TaurusDevicePanel label

cmft:
Avoid large fullname in the TaurusDevicePanel label.
Add tooltip to show the fullname.


cmft:
To check use:

```
import taurus

print taurus.Release.version
d1 = taurus.Device('sys/tg_test/1')
d2 = taurus.Device('tango://sys/tg_test/1')
d3 = taurus.Device('tango://pc253.cells.es:10000/sys/tg_test/1')
d4 = taurus.Device('tango://pc253:10000/sys/tg_test/1')

print d1 == d2
print d2 == d3
print d1 == d3
print d1 == d4
```

PR #776, opened 2018-07-23
Define correct focus on a 'apply' button

### LINKED ISSUE #772: TaurusPlotConfigDialog not intuitive UX
In TaurusPlotConigDialog after pressing "enter" the dialog will close and the user will lose all changes in the configuration. The more intuitive is that TaurusPlotConigDialog stores the configuration in this situation. We can achieve this in two ways. The first is to add an additional button with focus, for example the name "Ok" or change the focus from "Close" to "Apply". The second scenario has more impact on UX, because in this case we do not close the window. You can see this problem and suggestions for its solution in the screenshot. I realize that experienced Taurus users have become accustomed to this, but a lot of users in SOLARIS have reported this problem. What do you think?

SOLUTION 1:
solution1

SOLUTION 2:
solution2

KrystianKedron:
Fix the problem with different focus on a 'apply' and 'colse' in the TaurusPlotConfigDialog (depends of the X environment), by define this order in the Taurus. This bug cause different behavior after pressing 'enter' key. See #772.

Fixes #772

### DIFF
--- a/lib/taurus/qt/qtgui/plot/qwtdialog.py
+++ b/lib/taurus/qt/qtgui/plot/qwtdialog.py
@@ -182,11 +182,17 @@ class TaurusPlotConfigDialog(Qt.QDialog):
         else:
             self.ui.peaksComboBox.setCurrentIndex(3)
 
-        # define correct focus on a 'apply' button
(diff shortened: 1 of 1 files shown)


cpascual:
Tested. Merging (it will get into the Jul18 release). Thanks @KrystianKedron !

PR #777, opened 2018-07-24
Allow travis to fail on builds for debian 8

cpascual:
The debian8 (Jessie) builds in travis are currently getting a lot
of spurious "no output for 10m" failures.
This should be addressed properly but as a workaround, just mark
this build as "allowed to fail"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ env:
 
 matrix:
     allow_failures:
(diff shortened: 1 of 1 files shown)


reszelaz:
I agree. In sardana we also observed that recently the tests are spuriously failing more than usually. Now I will pay attention if these are only the jessie tests or also the stretch ones.

PR #778, opened 2018-07-24
Fix sphinx

cpascual:
Fix all sphinx warnings and set travis to fail if new sphinx warnings are introduced


reszelaz:
All the sphinx warnings are fixed! Thanks! I was going to accept it but the tests on stretch are failing. I think it is not related to this PR. Should we rebuild them? 

Alos, I think it would be nicer to split the documentation build in another job so it will be easier to see what is the reason of CI failure, if these were docs or testsuite, and navigate to the corresponding output.

PR #781, opened 2018-07-24
Make "Change Tango Host" action invisible

### LINKED ISSUE #379: TaurusMainWindow's "Change Tango Host" action does "nothing"
All `TaurusMainWindow `-based applications have a "Change Tango Host" Action by default (TaurusMainWindow menu-->Taurus-->Change Tango Host)

In its default implementation, it just stores the new tango host variable in a `tangoHost` property and does nothing else. Users may get confused by this (they may reasonably expect e.g. that all models in the whole GUI get updated to new values, like jive would do). 

But note forcing a change of all models is both technically difficult (think perspectives) and probably not desired in many situations.

IMHO, the only safe thing to add  to the *default* behaviour of this action is to emit a `tangoHostChanged` signal. Then, **for the particular case of TaurusGui aplications**, the MainWindow could be registered by default as a writer and reader of a shared data called "TangoHost". In this way, it would be up to the person adding a panel to register it as a reader/writer and act on changes of "TangoHost" as they see fit.

But note that even if implementing the above changes, the action would still "do nothing" in general, so IMHO it would make sense to either disable, or even hide it *by default* (i.e. so that it does not appear in menus). This would make special sense to me considering that the default options and actions should gradually be made less "Tango-centric".

cpascual:
The "taurus" menu in TaurusMainWindow contains the "Change Tango Host"
action, which is tango-centric, does nothing and is deprecated
(see #379). Remove the action from the menu.

Note that the action itself is not removed from the TaurusMainWindow. 
It is just made invisible in order not to change the API on a minor change.

Fix #379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ develop branch) won't be reflected in this file.
 ### Changed
 - Serialization mode now is explicitly set to Serial
   in the case of TangoFactory (Taurus defaults to Concurrent) (#678)
+- TaurusMainWindow's "Change Tango Host" action is now invisible (#781)
 
 ### Fixed
 - TaurusModel ignoring the serialization mode (#678)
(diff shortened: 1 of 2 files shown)


reszelaz:
Just to not forget, there is still the empty Taurus menu in the menu bar. In the future we could try to remove it.

For the moment merging..

PR #792, opened 2018-08-20
Handle exception when adding listener to non-ready device

### LINKED ISSUE #791: device model not set if GUI started before device server
Hello,
I have updated to taurus4.4.0. Not sure if the observation was there before, but here we go:
I have two TaurusCommandButtons. I set the model on these and the command. If I start the gui now before the device server, I see that only the button I set the model on first is correctly connected and executes its command. The other button however has no model and can therefore not execute any command. When started normally, both buttons work. When during normal operation I stop the device server and restart it, I also cannot see a problem there.
You can reproduce this with the TangoTest device server.
First run the following script with argument `sys/tg_test/1`:

```
import sys

from taurus.qt.qtgui.util.ui import UILoadable
from taurus.qt.qtgui.button import TaurusCommandButton
from taurus.qt.qtgui.base.taurusbase import TaurusBaseWidget, TaurusBaseWritableWidget

if __name__ == '__main__':

    instance = sys.argv[1]

    app = TaurusApplication(app_name='MyGui')
    
    panel = Qt.QWidget() 
    frame = Qt.QFrame(panel)
    frame.resize(200,200)
	
    button_1 = TaurusCommandButton(panel)
    button_1.setCommand("SwitchStates")
    button_1.setCustomText("Switch States")
    button_1.move(10,10)
    button_2 = TaurusCommandButton(panel)
    button_2.setCommand("DevBoolean")
    button_2.setParameters([1])
    button_2.setCustomText("DevBoolean")
    button_2.move(10,40)
    button_2.setModel(instance)
    button_1.setModel(instance)


    panel.show()
    app.exec_()
```

Then start the TangoTest device server. Push the 'Switch States' button and you should get an error message.
Thanks for your time and taking a look into this!

cpascual:
TangoDevice.addListener() fails to handle an exception when a second
listener is added and the device is not ready. Handle it.

Fixes #791

### DIFF
--- a/lib/taurus/core/tango/tangodevice.py
+++ b/lib/taurus/core/tango/tangodevice.py
@@ -281,7 +281,12 @@ class TangoDevice(TaurusDevice):
         if weWereListening:
             # We were listening already, so we must fake an event to the new
             # subscribed listener with the current value
-            evt_value = self.__decode(self.stateObj.read())
+            try:
+                evt_value = self.__decode(self.stateObj.read())
+            except:
+                # the value may not be available (e.g. if device is not ready)
+                self.debug('Cannot read state')
+                return ret
             listeners = hasattr(listener, '__iter__') and listener or [
                 listener]
             self.fireEvent(TaurusEventType.Change, evt_value, listeners)


cpascual:
Since we got a positive review from @cfalcon and confirmation from @CorF4 (in https://github.com/taurus-org/taurus/issues/791#issuecomment-414268303 ), I am merging this.

PR #794, opened 2018-08-21
Increase the minimal height for TaurusWheelEdit.

### LINKED ISSUE #788: TaurusWheelEdit does not claim enough vertical space in a form
The following snippet shows how the second item in the form (the TaurusWheelEdit is too constrained vertically: 

```python

if __name__ == '__main__':
    import sys
    app = TaurusApplication()
    w = TaurusForm()
    model = ['sys/tg_test/1/double_scalar'] * 10
    w.setModel(model)
    w[1].setWriteWidgetClass('TaurusWheelEdit')
    w.show()
    sys.exit(app.exec_())
```

image

cmft:
Sometimes TaurusWheelEdit is too constrained vertically.

Increase the minimal height for TaurusWheelEdit.

Fix #788

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusvalue.py
+++ b/lib/taurus/qt/qtgui/panel/taurusvalue.py
@@ -872,7 +872,11 @@ class TaurusValue(Qt.QWidget, TaurusBaseWidget):
             self._writeWidget.setForceDangerousOperations(
                 self.getForceDangerousOperations())
             if self.minimumHeight() is not None:
-                self._writeWidget.setMinimumHeight(self.minimumHeight())
+                if self.writeWidgetClassID == 'TaurusWheelEdit':
+                    self._writeWidget.setMinimumHeight(
+                        self.minimumHeight() * 2)
+                else:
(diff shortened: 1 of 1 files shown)


cpascual:
Hi, thanks for the PR, but I am not too keen on the approach for the fix for the following reasons:

- Handling this as a explicit special case at TaurusValue level seems too much of a hack. It should be the TaurusWheelEdit itself that provides a sensible minimumHeight
- I am not sure that `writeWidgetclassID` can be relied upon in all cases
- Why a factor of 2? Why not 1.8 or 2.1? Again it all looks a bit too hack-ish

cpascual:
I reply myself:
> first we need to check why the TV's minimum height is set to something other than None

The answer is in this "TODO" in the `TaurusForm.fillWithChildren()` implementation:

```
# @todo UGLY... See if this can be done in other ways... (this causes trouble with widget that need more vertical space , like PoolMotorTV)
            widget.setMinimumHeight(20)
```

So, IMHO, solving that TODO is the way to go... I am threfore closing this PR as rejected

PR #795, opened 2018-08-22
Set QWheel minimum height (Fix #788)

### LINKED ISSUE #788: TaurusWheelEdit does not claim enough vertical space in a form

cpascual:
Fix #788 by explicitly setting the minimum height of the QWheel widget
when it is being constructed.

This is a second attempt at fixing it after rejecting #794 approach.
Note that the ugly behaviour of TaurusForm of forcing minimum heights to its subwidgets (see https://github.com/taurus-org/taurus/pull/794#issuecomment-414935963) is not yet dealt-with (but it is not necessary to fix #788)

### DIFF
--- a/lib/taurus/qt/qtgui/input/qwheel.py
+++ b/lib/taurus/qt/qtgui/input/qwheel.py
@@ -325,6 +325,15 @@ class QWheelEdit(Qt.QFrame):
                                 self.getDecDigitCount())
         ed.setVisible(False)
         self._editor = ed
+
+        # set the minimum height for the widget
+        # (otherwise the hints seem to be ignored by the layouts)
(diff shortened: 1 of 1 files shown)


teresanunez:
I have just checked, as @cpascual wrote #788 is fixed, so IMO could be merged (and the remaining problem can be treated separately).

PR #798, opened 2018-08-27
Fix TaurusValueLineEdit ignores #wvalue.magnitude fragment

### LINKED ISSUE #797: TaurusValueLineEdit ignores #wvalue.magnitude fragment
There has been a regression and now the line edits are failing to act on `#wvalue.magnitude` fragments.

To reproduce:
- Configure sys/tg_test/1/short_scalar to have some units
- run:

```
python -m taurus.qt.qtgui.input.tauruslineedit --taurus-log-level=debug \
            'sys/tg_test/1/short_scalar#wvalue.magnitude' \
            'sys/tg_test/1/short_scalar'
```

This should show 2 lineedits, the first without units and the second with them.

The reason for the failure is an exception in `TaurusValueLineEdit.setValue`, as it can be seen in the debug message: 

`MainThread     DEBUG    2018-08-23 12:56:40,654 TaurusValueLineEdit.sys/tg_test/1/short_scalar#wvalue.magnitude: Cannot enforce fragment. Reason: AttributeError("'NoneType' object has no attribute 'units'",)
`

After some debugging, it seems that the validator is not initialized when setValue is called (which causes this exception).

cmft:
Line edits are failing to act on #wvalue.magnitude fragments.
The validator is not initialized when setValue  is called in
the postAttach method.

Fix #797 overwriting the postAttach method.

### DIFF
--- a/lib/taurus/qt/qtgui/input/tauruslineedit.py
+++ b/lib/taurus/qt/qtgui/input/tauruslineedit.py
@@ -105,6 +105,18 @@ class TaurusValueLineEdit(Qt.QLineEdit, TaurusBaseWritableWidget):
         """reimplement to avoid autoapply on every partial edition"""
         self.emitValueChanged()
 
+    def postAttach(self):
+        """reimplemented from :class:`TaurusBaseWritableWidget`"""
+        TaurusBaseWritableWidget.postAttach(self)
+        if self.isAttached():
+            try:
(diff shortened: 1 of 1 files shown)


cpascual:
Merged **after refactoring** (in collab with @cmft) in https://github.com/taurus-org/taurus/commit/b67ee85c2d54086b29918ed10a04d275fbc3da56

PR #800, opened 2018-08-28
Fix TaurusForm "Set Format" option does not work in compact mode

### LINKED ISSUE #787: TaurusForm "Set Format" option does not work in compact mode
To reproduce:

- run `taurusform sys/tg_test/1/ampli sys/tg_test/1/float_scalar`
- change the first item to compact
- go  to context menu -> set formatter (all items) --> enter `>>>>{}<<<<`

==> the first element (the compact one) ignored the new format

cmft:
The compact elements of a taurusform ignore the set format option.

Propagate the format when the read widget is compacted and vice versa.

Fix #787

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusform.py
+++ b/lib/taurus/qt/qtgui/panel/taurusform.py
@@ -397,9 +397,8 @@ class TaurusForm(TaurusWidget):
         """
         TaurusWidget.setFormat(self, format)
         for item in self.getItems():
-            rw = item.readWidget(followCompact=True)
-            if hasattr(rw, 'setFormat'):
-                rw.setFormat(format)
+            if hasattr(item, 'setFormat'):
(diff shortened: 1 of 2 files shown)


cpascual:
The proposed solution seems to fix the problem, but raises an exception when a compact widget is created from a read-only taurusvalue:

```python

if __name__ == '__main__':
    import sys
    app = TaurusApplication()

    w = TaurusForm()
    w.setModifiableByUser(True)

    model = ['eval:1',
             'sys/tg_test/1/ampli',
            ]

    w.setModel(model * 2)

    for item in w[len(model):]:
        item.setCompact(True)

    w.setFormat('>>>>{}<<<<')

    w.show()
    w.resize(600,455)
    sys.exit(app.exec_())
```

I am not sure if the solution is to clear the compact flag when the widget was not compacted or to handle the exception or....

PR #805, opened 2018-08-30
Fix Taurusform: Change label fails for devices

### LINKED ISSUE #786: Taurusform: Change label fails for devices
To reproduce:

- run `taurusform sys/tg_test/1 sys/tg_test/1/ampli`
- Go to  context menu --> change labels (all items) --> enter `{dev.name}`

==> the label changes ok for the attr but not for the device:

image

cmft:
Taurus form change label context menu option does not work
when the set model is a device URI.

Fix #786

### DIFF
--- a/lib/taurus/core/tango/tangodevice.py
+++ b/lib/taurus/core/tango/tangodevice.py
@@ -168,9 +168,11 @@ class TangoDevice(TaurusDevice):
             rvalue of the returned TangoAttributeValue is now a member of
             TaurusDevState instead of TaurusSWDevState
         """
+        if not cache:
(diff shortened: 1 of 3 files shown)


cpascual:
Here is a snippet for checking:

```python

if __name__ == '__main__':
    import sys
    app = TaurusApplication()

    w = TaurusForm()
    w.setModifiableByUser(True)

    model = ['sys/tg_test/1',
             'sys/tg_test/1/ampli',
             'eval:@foo',
             'eval:@foo/1',
             ]
    w.setModel(model * 2)

    for item in w[len(model):]:
        item.labelConfig = '{dev.fullname}'

    w.show()
    w.resize(600,455)
    sys.exit(app.exec_())
```

PR #808, opened 2018-09-14
Fix wrong subscription state

cmft:
If the FILTER_OLD_TANGO_EVENTS feature is enable the
tangoAttribute _pushAttrEvent method could keep the subscription
state and its management in a wrong state.

Move the event filter after the subscription management fix the
inconsistency.


cpascual:
After reviewing this together with @cfalcon, we decided that a better approach is to add a check of the subscription state in the filter

Issue #811, opened 2018-09-18
Remove alba-specific code from taurus.qt.qtgui.panel.report


cpascual:
`taurus.qt.qtgui.panel.report.albareport` is ALBA-specific code which should be removed from taurus.

PR #813, opened 2018-09-26
Email direction update

fbecheri:
The email "sendto" field has changed from controls@rt.cells.es to 
controls.desk@cells.es


cpascual:
This is a alba-specific code which is probably going to be removed in taurus4. See #811 
I am merging this into the support-3.x since it does not harm anybody else.

PR #819, opened 2018-10-09
Fix Qt API collision in taurusnexuswidget

cpascual:
running `cd <taurus>/lib/taurus/qt/qtgui/extra_nexus/;python -m taurusnexuswidget` fails
because the early import of PyMCA sets the Qt API to 1. Do the
taurus.external.qt import before to avoid it.


teresanunez:
IMO it can be merged.

PR #827, opened 2018-11-09
Fix taurusfactory.getAttribute method

cmft:
getAttribute method assumes the scheme for creating the device.
It fails when you do not use the default scheme.

Fix it, composing the device name URI with the tuple scheme
and devname.

### DIFF
--- a/lib/taurus/core/taurusfactory.py
+++ b/lib/taurus/core/taurusfactory.py
@@ -217,7 +217,8 @@ class TaurusFactory(object):
         try:
             # this works only if the devname is present in the attr full name
             # (not all schemes are constructed in this way)
-            devname = v.getUriGroups(fullname)['devname']
+            groups = v.getUriGroups(fullname)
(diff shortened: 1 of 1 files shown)


lib/taurus/core/taurusfactory.py
```diff
@@ -217,7 +217,8 @@ def getAttribute(self, name):
         try:
             # this works only if the devname is present in the attr full name
             # (not all schemes are constructed in this way)
-            devname = v.getUriGroups(fullname)['devname']
+            groups = v.getUriGroups(fullname)
+            devname = '{0}:{1}'.format(groups['scheme'], groups['devname'])
```
cpascual: I am not sure this is the right way to deal with this: in principle, XXXFactory.getDevice (and other methods for getting Attribute, Database, etc.) should assume that the scheme is the one of the XXXFactory itself (so it should not require thatthe scheme is passed explicitly)

cmft: Agree. It seems it is not need

PR #829, opened 2018-11-15
(py3) Fix problems with qbase models

cpascual:
To reproduce, compare the behaviour of this snippet with py2 (ok) and py3 (fails):

```python
from taurus.qt.qtgui.tree import TaurusDbTreeWidget
from taurus.core import TaurusElementType
from taurus.core.tango import TangoDatabase

if __name__ == '__main__':
    import sys
    print('PYVERSION', sys.version)
    app = TaurusApplication()

    w = TaurusDbTreeWidget(perspective=TaurusElementType.DeviceClass)

    model = TangoDatabase().fullname
    w.setModel(model)

    w.show()
    sys.exit(app.exec_())
```


mrosanes:
Code reviewed and tested.
Tested with python2.7 on local machine it works correctly.
Tested with python 3.5.3 on docker image taurus-stretch-py3 where it works correctly as well.

PR #831, opened 2018-12-05
Make TangoAttrValue copyable

cpascual:
TangoAttrValue implements `__getattr__` to map some API from TangoAttribute and PyTango values. But the current implementation makes copy.copy() on a TangoAttrValue to fail because some special methods such as `__getstate__` are also mapped (which is not ok).

This can be reproduced with: 

```python
import taurus
import copy
v = taurus.Attribute('sys/tg_test/1/double_scalar').read()  # <-- TangoAttributeValue => FAILS
# v = taurus.Attribute('eval:1').read()  # <-- EvaluationAttrValue => OK
print(v)
w = copy.copy(v) 
```
The proposed solution is to limit the mapping of `__getattr__` to non-special methods.

Note: this was reported by @hayg25 in https://github.com/taurus-org/taurus_pyqtgraph/issues/3

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -140,7 +140,16 @@ class TangoAttrValue(TaurusAttrValue):
         self.quality = quality_from_tango(p.quality)
 
     def __getattr__(self, name):
+        """
+        If the member `name` is not defined in this class, try to get it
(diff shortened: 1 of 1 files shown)


cpascual:
@taurus-org/integrators , can you please check and integrate this PR?

cpascual:
> iMO it can be integrated.

feel free to merge it yourself then!  ;)

PR #833, opened 2018-12-17
Add scheme entry point in TaurusManager

cmft:
Add entry point access `taurus.core.schemes` as another option
to register schemes in Taurus.

TaurusManager._get_plugin_classes method add this code as
experimental feature


cpascual:
Merging. Thanks!

PR #835, opened 2018-12-20
Fix #834-Remove problematic findCaller implementation

### LINKED ISSUE #834: ThreadPool not able to report traceback
One can use the following snippet to reproduce the problem:
```python
import time
from taurus.core.util.threadpool import ThreadPool

pool = ThreadPool()

def job():
    1/0

pool.add(job)
time.sleep(1)
```
In support-3.x:
```
ThreadPool.W001 ERROR    2018-12-18 16:46:14,679 ThreadPool.ThreadPool.W001: Uncaught exception running job 'job' called from thread MainThread:
  File "/home/zreszela/.PyCharm2018.2/config/scratches/scratch_2.py", line 9, in <module>
    pool.add(job)
  File "/home/zreszela/workspace/taurus/lib/taurus/core/util/threadpool.py", line 134, in run
    cmd(*args, **kw)
  File "/home/zreszela/.PyCharm2018.2/config/scratches/scratch_2.py", line 7, in job
    1/0
ZeroDivisionError: integer division or modulo by zero
```

In develop:
```
ThreadPool.W001 ERROR    2018-12-18 16:47:18,487 ThreadPool.ThreadPool.W001: Uncaught exception running job 'job' called from thread MainThread:
  File "/home/zreszela/.PyCharm2018.2/config/scratches/scratch_2.py", line 9, in <module>
    pool.add(job)
  File "/usr/lib/python2.7/logging/__init__.py", line 1277, in _log
    fn, lno, func = self.findCaller()
ValueError: too many values to unpack
```

cpascual:
Fix #834:

`taurus.core.util.log` reimplements `logging.Logger` class to overwrite
the findCaller method. This reimplementation is not simultaneously
py2+py3 compatible and it is almost identical to the py3
implementation, breaking things on py2.
Remove our reimplementation and use the logging ones (py2 or py3)
instead.

### DIFF
--- a/lib/taurus/core/util/log.py
+++ b/lib/taurus/core/util/log.py
@@ -414,41 +414,6 @@ class LogExceptHook(BaseExceptHook):
         self._log.log(self._level, "Unhandled exception:\n%s", text)
 
 
-class _Logger(logging.Logger):
-
-
-    def findCaller(self, stack_info=False):
-        """
-        Find the stack frame of the caller so that we can note the source
-        file name, line number and function name.
-        """
-        f = currentframe()
-        # On some versions of IronPython, currentframe() returns None if
-        # IronPython isn't run with -X:Frames.
-        if f is not None:
-            f = f.f_back
-        rv = "(unknown file)", 0, "(unknown function)", None
-        while hasattr(f, "f_code"):
-            co = f.f_code
-            filename = os.path.normcase(co.co_filename)
-            if filename in (_srcfile, logging._srcfile):
-                f = f.f_back
-                continue
-            
-            sinfo = None
-            if stack_info:
-                sio = io.StringIO()
-                sio.write('Stack (most recent call last):\n')
-                traceback.print_stack(f, file=sio)
-                sinfo = sio.getvalue()
-                if sinfo[-1] == '\n':
-                    sinfo = sinfo[:-1]
-                sio.close()
-            rv = (co.co_filename, f.f_lineno, co.co_name, sinfo)
-            break
-        return rv
-
-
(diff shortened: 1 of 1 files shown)


cpascual:
Ouch....

So now we should decide how we consider this PR: Should it be consider a change of API that breaks backwards compatibility, or a bug fix that fixes a wrong implementation?

My opinion is not strong here, but I tend to consider the latter because:
According to the docs, the `Logger.findCaller` method should return a [3-tuple in python 2.7](https://docs.python.org/2/library/logging.html#logging.Logger.findCaller) but [a 4-tuple in python 3](https://docs.python.org/3/library/logging.html#logging.Logger.findCaller)... and our own reimplementation did not specify anything different in its docstring. Therefore I consider that our previous implementation (returning always a 4-tuple) was actually buggy.

So, IMHO, we should **not** revert this PR but we should instead change the code using it to be aware of this difference between the py2 and py3 implementations. This can be easily done by searching and replacing `findCaller()` by `findCaller()[:3]`


**BUT** The only thing that seems strange to me is the fact that in the exception that you are reporting, the problematic line is expecting 3 values and apparently `findCaller` is returning 4 (even if it is using python2.7), which is precisely what I would **not** expect from the latest taurus!  
Can you please re-check that your taurus version is indeed the one with this PR applied?

reszelaz:
Thanks for the explanation! I agree with the strategy on how to deal with that.
Unfortunatelly, I'm not able to reproduce it now. I'm pretty sure I was using the latest taurus in the Docker container where I reproduced it yesterday night.. I will keep an eye on it and eventually come back.

Issue #848, opened 2019-01-27
Infinite recursion when Copying EvaluationAttrValue


asenchenko:
While adding python3 support to [taurus_pyqtgraph](https://github.com/taurus-org/taurus_pyqtgraph) i found that coping EvaluationAttrValue ( using copy module ) rises exception:
```
Traceback (most recent call last):
  File "taurustrendset.py", line 62, in <module>
    c2.setModel('eval:rand(2)')
  File "/home/user/taurus_pyqtgraph/taurus_pyqtgraph/taurustrendset.py", line 132, in setModel
    self._forceRead()
  File "/home/user/taurus_pyqtgraph/taurus_pyqtgraph/taurustrendset.py", line 340, in _forceRead
    value = copy.copy(value)
  File "/home/user/venv3/lib/python3.6/copy.py", line 106, in copy
    return _reconstruct(x, None, *rv)
  File "/home/user/venv3/lib/python3.6/copy.py", line 281, in _reconstruct
    if hasattr(y, '__setstate__'):
  File "/home/user/venv3/lib/python3.6/site-packages/taurus-4.4.1a0-py3.6.egg/taurus/core/evaluation/evalattribute.py", line 64, in __getattr__
    ret = getattr(self._attrRef, name)
  File "/home/user/venv3/lib/python3.6/site-packages/taurus-4.4.1a0-py3.6.egg/taurus/core/evaluation/evalattribute.py", line 64, in __getattr__
    ret = getattr(self._attrRef, name)
  File "/home/user/venv3/lib/python3.6/site-packages/taurus-4.4.1a0-py3.6.egg/taurus/core/evaluation/evalattribute.py", line 64, in __getattr__
    ret = getattr(self._attrRef, name)
  [Previous line repeated 327 more times]
RecursionError: maximum recursion depth exceeded while calling a Python object
```
Copy module uses `__copy__` method. If this method does not exist, it  falls back to pickle protocol (`__reduce__, __setstate__, __getstate__ etc.`).  `_reconstruct` function from copyreg module is used to create empty object (internally calls `__new__`).  After creating object `copy` checks, is there `__setstate__` method. `hasattr` invokes `__getattr__`. But there is no `_attrRef`, so it recursively invokes `__getattr__` again and again.  
For more details, see [copy.py](https://github.com/python/cpython/blob/3.6/Lib/copy.py) and [copyreg.py](https://github.com/python/cpython/blob/3.6/Lib/copyreg.py).

[...]

PR #849, opened 2019-01-27
Allow copying of EvaluationAttrValue (Fixes issue #848).

asenchenko:
Fixes issue #848


cpascual:
Hi, thanks a lot for the proposed fix. 

However, note that this is very similar to what we did for the `TangoAttrValue` object in #831. In that case we were a bit more broad and applied this same recipe to all methods starting and ending in `__`. I personally prefer that broader solution, so I am substituting it here before merging.

cpascual:
I'll merge this as soon as travis stops failing spuriously and gives me a green light (see #851).

PR #854, opened 2019-01-30
Fix Tango authority did not manage connection failures

### LINKED ISSUE #853: Tango authority did not manage connection failures
In taurus if the Tango authority is not reachable after 5 seconds  it returns None. 
In case the authority did not exist it returns None straight away . 
To be coherence with the other models it should be return a taurus object instead.

```
import taurus
auth = taurus.Authority('tango://foo:10000')
```

cmft:
If the tango database did not exist or it can not be reacheable
the TangoAuthority returns None instead of an object.

Fix it, catching possible exceptions during PyTango Database creation.

Fix #853

### DIFF
--- a/lib/taurus/core/tango/tangodatabase.py
+++ b/lib/taurus/core/tango/tangodatabase.py
@@ -683,7 +683,8 @@ class TangoAuthority(TaurusAuthority):
             self.dbObj = Database(host, port)
         except Exception as e:
             from taurus import warning
-            warning(e)
+            warning("Cannot connect with Tango DB: %r", e)
+            self.dbObj = None
 
         self._dbProxy = None
         self._dbCache = None
--- a/lib/taurus/core/tango/tangodatabase.py
+++ b/lib/taurus/core/tango/tangodatabase.py
@@ -679,8 +679,12 @@ class TangoAuthority(TaurusAuthority):
 
         # Set host to fqdn
         host = socket.getfqdn(host)
+        try:
+            self.dbObj = Database(host, port)
+        except Exception as e:
+            from taurus import warning
+            warning(e)
 
-        self.dbObj = Database(host, port)
         self._dbProxy = None
         self._dbCache = None
 


lib/taurus/core/tango/tangodatabase.py
```diff
@@ -679,8 +679,12 @@ def __init__(self, host=None, port=None, parent=None):
 
         # Set host to fqdn
         host = socket.getfqdn(host)
+        try:
+            self.dbObj = Database(host, port)
+        except Exception as e:
+            from taurus import warning
```
cpascual: Since TangoAuthority is a logger object, why not use `self.warning`?

lib/taurus/core/tango/tangodatabase.py
```diff
@@ -679,8 +679,12 @@ def __init__(self, host=None, port=None, parent=None):
...
+            warning(e)
```
cpascual: IMHO, the warning message should be a bit more descriptive. For example:
`warning("Cannot connect with Tango DB: %r", e)

lib/taurus/core/tango/tangodatabase.py
```diff
@@ -679,8 +679,12 @@ def __init__(self, host=None, port=None, parent=None):
 
         # Set host to fqdn
         host = socket.getfqdn(host)
+        try:
+            self.dbObj = Database(host, port)
```
cpascual: If a exception occurs,  `self.dbObj` is not defined. I would set it to `None` instead

lib/taurus/core/tango/tangodatabase.py
cmft: Base class is not initialized yet

lib/taurus/core/tango/tangodatabase.py
cmft: Agree with it

lib/taurus/core/tango/tangodatabase.py
cmft: I change it

cpascual:
After some conversations with @cmft , we are unsure about whether this is a good idea as it is now:
in the proposed implementation, if the tango DB is not available, the factory will return a useless (not connected) TangoAuthority object **and will keep it in its cache**, so further attempts to retrieve it again will always be given the disconnected object... even if the Tango DB actually became available again.

So, I think that it is better to either:
a) keep things as they were (returning return None and not caching the disconnected object), or
b) implement a polling mechanism in the `TangoAuthority` object so that it periodically tries to reconnect and reinitialize itself and notifies any affected device and attributes in case a reconnection succeeds (this would be nice, but IMHO it may add too much complexity for a small gain in a corner case)

Other people, @taurus-org/integrators , ... any thoughts?

reszelaz:
> in the proposed implementation, if the tango DB is not available, the factory will return a useless (not connected) TangoAuthority object and will keep it in its cache, so further attempts to retrieve it again will always be given the disconnected object... even if the Tango DB actually became available again.

Good reasoning!

> So, I think that it is better to either:

From the options you listed I would choose a).

But maybe an intermediate option would be something like:

Return an autority object without connection to the Tango database and whenever someone tries to use this object and it requires communication with the database it will try to establish a connection. If possible - fine, if not possible - exception.

Implementation could be done with  adecorator which would mark the API requiring communication with the Tango database.

cpascual:
I am closing this. The issue #853 is left open as a "TODO"

PR #855, opened 2019-02-01
Print exception message only instead of the whole repr when error while loading settings

reszelaz:
This is a cosmetic and trivial change. I think it helps in the readability of the error message.


cpascual:
I merged after doing a PEP8 correction and testing it.

Thanks!

PR #856, opened 2019-02-01
Replace CLI scripts with a single one

cpascual:
It would be nice to replace the current console scripts of taurus by a single one that used subcommands (similar to e.g. `git`).
For example, instead of `taurusform` or `taurusdemo` we could have `taurus form` and `taurus demo`.

The advantages are:

- less pollution of binary namespace
- simplified packaging and documentation (just one manpage to maintain)
- clearer, central point of access for all taurus related commands
- possibility of defining an entry point so that plugins also provide their own subcommands with a unified interface


cpascual:
I pushed another implementation, based on [click](https://click.palletsprojects.com) (thanks to @tiagocoutinho !) .

For now it implements the `taurus` command with 2 subcommands (`form` and `demo`) as a proof of concept. Besides,https://github.com/taurus-org/taurus_pyqtgraph/pull/24 also demonstrates how a plugin can provide its own subcommands

stanislaw55:
Hi,
have you considered sharing common options between commands? It could be achieved by something similar to [this](https://stackoverflow.com/a/40195800) SO answer. 
If you want, I can provide code snippets (not sure about how to contribute to this PR. Make PR to your branch? Or just leave comments with code snippets?)

cpascual:
Hi @stanislaw55 , thanks for the suggestion!

Indeed I was considering creating some common options (e.g. the `models` argument). The approach that I was considering is similar to the one you propose, and was inspired by the "best practices" recommended [here](https://github.com/click-contrib/click-plugins#best-practices-and-extra-credit).

I also want to remove all usages of the `taurus.core.util.argparse` module (which depends on optparse) from the taurus core. This involves refactoring ~20 `main` or `demo` functions scattered in taurus, and some of them could even become subcommands of `taurus` or subcommands of `taurus demo`.

But in order to not delaying this PR any longer, my plan was to propose it for integration as it is now (IMHO it is ready for integration since all the old console-scripts are now implemented as subcommands) and then go for the mentioned improvements in a subsequent iteration. This would also simplify the way for you to contribute to it: you could do a PR to develop implementing the common options once this PR is merged. Is this ok for you?

stanislaw55:
Thanks for answer,
Yes, that way it could  be easier to contribute. I'll just wait for this PR to be merged and then provide another one with this common options

cpascual:
Hi @taurus-org/integrators  , can you review/integrate this?

@stanislaw55, since you are already involved, it would be great if you could also review it (the integrators would appreciate your feedback).

cpascual:
BTW: note that the old commands are still provided by the [taurus_legacy_cli plugin package](https://github.com/taurus-org/taurus_legacy_cli), in case they are required for backwards compatibility

cpascual:
I've added 3 last-minute commits (mostly doc-related)

cpascual:
I created 3 Issues as follow ups from this PR: #922, #923, #924

teresanunez:
Hi @cpascual, I have just tested this and also the taurus_legacy_ci plugin package (to check the
 forward compatibility). IMO it can be merged. I can do it if nobody is against it.

cpascual:
Thanks @teresanunez for the review!
@cfalcon is also looking at it at this moment, so I propose to wait for his feedback and then go ahead with the merge.

cmft:
Minor issues
- taurus newgui command does not work 
- I will rename taurus dev command by taurus device

cpascual:
I am committing fixes for the minor issues from @cmft  plus an enhancement to reduce log spam when running the taurus command

Also, I tested the autocompletion feature which works well if activated following instructions in https://click.palletsprojects.com/en/7.x/bashcomplete/

stanislaw55:
Mino issues:
 - in taurus designer, `--taurus-path` shows `TEXT` in help message. Previously it was `TAURUSPATH`
 - `taurus --help` (and hence all other subcommands) reports deprecation warning about `Qwt5`. It is not in previous version. Is it caused by how groups are added to click command line launcher?

cpascual:
Thanks for the review:

>  - in taurus designer, `--taurus-path` shows `TEXT` in help message. Previously it was `TAURUSPATH`

I just pushed a fix for this

 > - `taurus --help` (and hence all other subcommands) reports deprecation warning about `Qwt5`. It is not in previous version. Is it caused by how groups are added to click command line launcher?

This should be already fixed (in 9bfc0e1 ) Can you please pull the latest changes and re-test this?

> autocompletion would be very nice but I think it can be another issue and PR, just like with shared option

The good news is that it is already done. Click provides it. You just need to enable it in your system following the instructions given [here](https://click.palletsprojects.com/en/7.x/bashcomplete/).
For example, in my debian stretch system, using bash, it is:

For enabling it system-wide:
```console
_TAURUS_COMPLETE=source taurus > taurus-complete.sh
sudo mv taurus-complete.sh  /etc/bash_completion.d/taurus
```

Alternatively, enabling it just for the current user
```console
_TAURUS_COMPLETE=source taurus > $HOME/.local/bin/taurus-complete.sh
echo '. $HOME/.local/bin/taurus-complete.sh' >> $HOME/.bashrc
```

Note: in the next debian package, I intend to configure it so that it is automatically enabled system-wide

cmft:
I found an issue with taurus gui foo (does not work) where foo is a non installed taurus gui

cpascual:
@cmft  :
> I found an issue with taurus gui foo (does not work) where foo is a non installed taurus gui

I just tried the following:
- created a GUI called foo with `taurus gui newgui` (this creates a `foo` dir with a `tgconf_foo` subdir)
- changed to dir `foo`
- run `taurus gui foo` 
  -it fails with `MainThread     ERROR    2019-06-19 12:11:21,503 TaurusGui: cannot import foo or tgconf_foo`
- run `taurus gui tgconf_foo` . It works well.

So, I agree there is a problem (at the very least with the error message, which suggests that it tried tgconf_foo)

**BUT**, I checked and I get exactly the same behaviour in the develop branch if using the old `taurusgui` script.

So IMHO, this can be reported as a bug and not block the integration of this PR

cpascual:
I think that all the issues are cleared for integrating this.
@teresanunez , can you please merge it?

PR #860, opened 2019-02-07
Avoid redundant display of units in TaurusForm (Fixes #642)

### LINKED ISSUE #642: TaurusForm: Avoid displaying redundant units in write widget
When displaying a numeric scalar value in TaurusForm, the units are shown both in the write value widget and in the units widget:

image

While it is fine that the write widget accepts input in any compatible unit, the unit should not be shown when it is the same as the one displayed in the units widget.

Perhaps the write widget should transform its displayed text (when applying) to match the units shown by the units widget

cpascual:
TaurusForm uses TaurusValueLineEdit as a default write widget.
If the attribute is a numeric scalar, this shows the magnitude and the
units, which is redundant since the units are already displayed in the
units widget. Replace the default line edit in TarusValue by a
customized version that displays only the magnitude.

This is a (quick and dirty) fix for #642

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusvalue.py
+++ b/lib/taurus/qt/qtgui/panel/taurusvalue.py
@@ -204,6 +204,15 @@ class CenteredLed(TaurusLed):
     DefaultAlignment = Qt.Qt.AlignHCenter | Qt.Qt.AlignVCenter
 
 
+class UnitLessLineEdit(TaurusValueLineEdit):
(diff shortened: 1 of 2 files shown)


cpascual:
@taurus-org/integrators , can someone please review this ASAP?

If possible I would like to add it to the Jan19 release

cmft:
I reviewed and  tested the code and everything seems ok for me.

reszelaz:
Thanks @cmft for the tests. I'm merging..

PR #869, opened 2019-02-15
Taurus model selector

cmft:
This PR add new base classes to Taurus: TaurusModelSelector and TaurusModelSelectorItem
And implement the TangoModelSelectorItem specialization.

The TaurusModelSelector class defines an entry point where the ModelSelectorItem specialization can be registered via entry point (EP).

This EP is described in TEP13: https://github.com/taurus-org/taurus/blob/develop/doc/source/tep/TEP13.md#the-taurusmodelchooser-widget-support-for-non-tango-schemes

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusmodelchooser.py
+++ b/lib/taurus/qt/qtgui/panel/taurusmodelchooser.py
@@ -43,7 +43,9 @@ from taurus.core.util.containers import CaselessList
 from .taurusmodellist import TaurusModelList
 
 __all__ = ["TaurusModelSelectorTree", "TaurusModelChooser",
-           "TaurusModelSelector", "TaurusModelSelectorItem"]
+           "TaurusModelSelector", "TaurusModelSelectorItem",
+           "TangoModelSelectorItem"]
+
 
 class TaurusModelSelector(Qt.QTabWidget):
     """TaurusModelSelector is a QTabWidget container for
@@ -134,10 +136,13 @@ class TaurusModelSelectorItem(TaurusWidget):
     # Reimplement this property
     default_model = property(fget=_get_default_model, fset=_set_default_model)
 
+
+class TaurusModelSelectorTree(TaurusModelSelectorItem):
+
     addModels = Qt.pyqtSignal('QStringList')
 
     def __init__(self, parent=None, selectables=None, buttonsPos=None, designMode=None):
-        TaurusWidget.__init__(self, parent)
+        TaurusModelSelectorItem.__init__(self, parent)
         if selectables is None:
             selectables = [taurus.core.taurusbasetypes.TaurusElementType.Attribute, taurus.core.taurusbasetypes.TaurusElementType.Member,
                            taurus.core.taurusbasetypes.TaurusElementType.Device]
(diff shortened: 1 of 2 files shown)


lib/taurus/qt/qtgui/panel/taurusmodelchooser.py
```diff
@@ -29,24 +29,120 @@
...
+                taurus.warning(err)
+        # ---------------------------------------------------------------------
+
+    def __setTabItemModel(self):
+        w = self.currentWidget()
+        if w.model == "":
+            self.setCursor(QtCore.Qt.WaitCursor)
+            print(w)
+            w.setModel(w.default_model)
+            self.setCursor(QtCore.Qt.ArrowCursor)
```
cpascual: There are several things I would improve in `__setTabItemModel`:
- Do no not assume it  has a model (I know that right now you assume that it is a TaurusBaseComponent, but this limitation is probably not really needed)
- Consider the possibility of model being `None`
- Do not  assume that the cursor was an arrow before.  Instead, store the actual cursor before changing it to WaitCursor and restore it. 
- Do not do a print.
- It should be protected agains exceptions when setting the model 
Here is my proposed alternative:

```python
def __setTabItemModel(self):
    w = self.currentWidget()
    c = self.cursor()
    try:
        if not w.model:
            self.setCursor(QtCore.Qt.WaitCursor)
            w.setModel(w.default_model)
    except Exception as e:
        taurus.warning('Problem setting up selector: %r', e)
    finally:
        w.setCursor(c)
```

lib/taurus/qt/qtgui/panel/taurusmodelchooser.py
```diff
@@ -29,24 +29,120 @@
...
+from taurus.external.qt import Qt, QtCore
 import taurus.core
 from taurus.qt.qtgui.container import TaurusWidget
 from taurus.qt.qtgui.tree import TaurusDbTreeWidget
 from taurus.core.util.containers import CaselessList
 from .taurusmodellist import TaurusModelList
 
+__all__ = ["TaurusModelSelectorTree", "TaurusModelChooser",
+           "TaurusModelSelector", "TaurusModelSelectorItem",
+           "TangoModelSelectorItem"]
```
cpascual: Since the location of the `TangoModelSelectorItem` is not yet definitive, maybe we could skip including it in `__all__` for now, and just register it for the entrypoint using the whole module path in `<taurus>/setup.py`

setup.py
```diff
@@ -111,6 +111,8 @@ def get_release_info():
 ]
 
 entry_points = {'console_scripts': console_scripts,
+                'taurus.qt.qtgui.panel.TaurusModelSelector.items':
+                    ['Tango = taurus.qt.qtgui.panel:TangoModelSelectorItem',]
```
cpascual: See previous comment about not exporting `TangoModelSelectorItem` in  `taurus.qt.qtgui.panel` and use instead:

 `'Tango = taurus.qt.qtgui.panel.taurusmodelchooser:TangoModelSelectorItem'`

lib/taurus/qt/qtgui/panel/taurusmodelchooser.py
cmft: Done

setup.py
cmft: Done

lib/taurus/qt/qtgui/panel/taurusmodelchooser.py
cmft: Done

PR #874, opened 2019-02-28
Add lock in clearEventSet

mrosanes:
Protect clearEventSet method from class AttributeEventWait, using
the object's lock.


reszelaz:
Finally we don't need it in Sardana, cause the place where we introduced the call to `clearEventSet` in https://github.com/sardana-org/sardana/pull/1069 is alredy protected by lock.

PR #877, opened 2019-03-06
(doc) Add note in the taurus coding conventions about Qt imports

cpascual:
Just adding a note copied from TEP18 to clarify the policy about importing Qt


cpascual:
@taurus-org/integrators  (bump)  ;)

cpascual:
@reszelaz 
>The TEP18 link is broken.

Fixed

> If not we just point to TEP18 and in parenthesis indicate the header (How to import Qt modules when writing taurus code).

Given that the full text is already quoted just below, leaving the link to TEP18 should be enough (i.e., no need to point to the exact header),  IMHO

reszelaz:
Ok, merging now. Thanks!

PR #881, opened 2019-03-13
Adapt TaurusGraphicsAttributeItem to be compatible with python3

cmft:
Adapt QGraphicsTextBoxing and TaurusGraphicsAttributeItem classes
to be compatible with py2/py3.

### DIFF
--- a/lib/taurus/qt/qtgui/graphic/taurusgraphic.py
+++ b/lib/taurus/qt/qtgui/graphic/taurusgraphic.py
@@ -867,12 +867,18 @@ class QGraphicsTextBoxing(Qt.QGraphicsItemGroup):
     _TEXT_RATIO = 0.8
 
     def __init__(self, parent=None, scene=None):
-        Qt.QGraphicsItemGroup.__init__(self, parent, scene)
-        self._rect = Qt.QGraphicsRectItem(self, scene)
+        Qt.QGraphicsItemGroup.__init__(self, parent)
+        if scene is not None:
+            scene.addItem(self)
+        self._rect = Qt.QGraphicsRectItem(self)
+        if scene is not None:
+            scene.addItem(self._rect)
         self._rect.setBrush(Qt.QBrush(Qt.Qt.NoBrush))
         self._rect.setPen(Qt.QPen(Qt.Qt.NoPen))
-        self._text = Qt.QGraphicsTextItem(self, scene)
-        self._text.scale(self._TEXT_RATIO, self._TEXT_RATIO)
+        self._text = Qt.QGraphicsTextItem(self)
+        if scene is not None:
+            scene.addItem(self._text)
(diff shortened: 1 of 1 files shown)


lib/taurus/qt/qtgui/graphic/taurusgraphic.py
```diff
@@ -867,12 +867,18 @@ class QGraphicsTextBoxing(Qt.QGraphicsItemGroup):
...
+        if scene is not None:
+            scene.addItem(self._rect)
         self._rect.setBrush(Qt.QBrush(Qt.Qt.NoBrush))
         self._rect.setPen(Qt.QPen(Qt.Qt.NoPen))
-        self._text = Qt.QGraphicsTextItem(self, scene)
-        self._text.scale(self._TEXT_RATIO, self._TEXT_RATIO)
+        self._text = Qt.QGraphicsTextItem(self)
+        if scene is not None:
+            scene.addItem(self._text)
+        # self._text.scale(self._TEXT_RATIO, self._TEXT_RATIO)
```
cpascual: This line is now commented... can it really be safely ignored?

lib/taurus/qt/qtgui/graphic/taurusgraphic.py
```diff
@@ -1164,7 +1170,7 @@ def updateStyle(self):
             else:
                 _frName = 'rvalue.magnitude'
             text = self._currText = self.getDisplayValue(fragmentName=_frName)
-        self._currText = text.decode('unicode-escape')
+        self._currText = text
```
cpascual: We need to check if the decode can safely be removed...

lib/taurus/qt/qtgui/graphic/taurusgraphic.py
cmft: It was commented by mistake

cmft: Sorry, it must be changed since it fails with py3

lib/taurus/qt/qtgui/graphic/taurusgraphic.py
cmft: this line is decoding a str (return of `self.getDisplayValue()`  ) to unicode str
in python3 str does not have decode and the strings are unicode. Should be enough importing str from builtins and doing a cast?

cpascual:
I need to check it carefully...

cpascual:
I've checked with:

```console
python -m taurus.qt.qtgui.graphic.jdraw.jdraw_view ~/src/taurus/lib/taurus/qt/qtgui/graphic/jdraw/test/res/SimpleScalarViewer.jdw
```
With the **develop** version, this works only for py2+qt4 but it fails for py3+qt4, py3+qt5, py2+qt5
... while with this PR, it works for all combinations, so I am merging.

Thanks!

PR #883, opened 2019-03-18
Fix bytes/str issues in dropMimeData methods

cpascual:
Various implementations of dropMimeData in Taurus do casts of
QBytesArray to str which may be dangerous in py3. Use bytes and decode
instead.


cpascual:
closed as duplicate of #886

PR #886, opened 2019-03-19
Fix bytes/str issues in dropMimeData methods

cpascual:
Various implementations of dropMimeData in Taurus do casts of
QBytesArray to str which may be dangerous in py3. Use bytes and decode
instead.

This fixes , for example, the problem that can be reproduced by opening a model chooser and trying to drag&drop attributes from the tree to the list (it works in py2 but not in py3)

### DIFF
--- a/lib/taurus/qt/qtgui/extra_guiqwt/curvesmodel.py
+++ b/lib/taurus/qt/qtgui/extra_guiqwt/curvesmodel.py
@@ -30,6 +30,7 @@ curvesmodel Model and view for new CurveItem configuration
 from __future__ import print_function
 from builtins import next
 from builtins import str
+from builtins import bytes
 from builtins import range
 from builtins import object
 
@@ -320,11 +321,12 @@ class TaurusCurveItemTableModel(Qt.QAbstractTableModel):
             else:
                 column = parent.columnCount()
         if data.hasFormat(TAURUS_ATTR_MIME_TYPE):
-            self.setData(self.index(row, column),
-                         value=str(data.data(TAURUS_ATTR_MIME_TYPE)))
+            model = bytes(data.data(TAURUS_ATTR_MIME_TYPE)).decode("utf-8")
+            self.setData(self.index(row, column), value=model)
             return True
         elif data.hasFormat(TAURUS_MODEL_LIST_MIME_TYPE):
-            models = str(data.data(TAURUS_MODEL_LIST_MIME_TYPE)).split()
+            d = bytes(data.data(TAURUS_MODEL_LIST_MIME_TYPE))
+            models = d.decode("utf-8").split()
             if len(models) == 1:
                 self.setData(self.index(row, column), value=models[0])
                 return True
--- a/lib/taurus/qt/qtgui/panel/taurusmodellist.py
+++ b/lib/taurus/qt/qtgui/panel/taurusmodellist.py
@@ -27,6 +27,7 @@
 itemsmodel Model and view for new CurveItem configuration
 """
 from builtins import object
+from builtins import bytes
 #raise UnimplementedError('Under Construction!')
 
 import copy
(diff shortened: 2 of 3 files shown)


cpascual:
Hi @amilan  I think you already did tests with py3, could you please review this for integration?

reszelaz:
The code looks ok if we assume that the bytes will arrive in `utf-8` encoding. I was looking for this being documented, for example [here](https://www.riverbankcomputing.com/static/Docs/PyQt5/gotchas.html#python-strings-qt-strings-and-unicode) but could not find it explicitly said.

I also tested it with the taurus model chooser (when launching taurusform without model passed in cmd line) and it indeed fixes the problem. 

I couldn't find a way of testing the other two changes comming from `extra_guiqwt`. I thought to test it with this:

```
zreszela@pc255:~/workspace/taurus (develop)> python3 -m taurus.qt.qtgui.extra_guiqwt.plot sys/tg_test/1/wave
MainThread     INFO     2019-03-27 19:06:54,093 TaurusRootLogger: Using PyQt4 (v4.11.4 with Qt 4.8.7 and Python 3.5.3)
MainThread     INFO     2019-03-27 19:06:54,351 taurus.qt.qtgui.icon.icon: Setting Tango icon theme (from /home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/icon/)
/usr/lib/python3.5/runpy.py:125: RuntimeWarning: 'taurus.qt.qtgui.extra_guiqwt.plot' found in sys.modules after import of package 'taurus.qt.qtgui.extra_guiqwt', but prior to execution of 'taurus.qt.qtgui.extra_guiqwt.plot'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
MainThread     WARNING  2019-03-27 19:06:55,014 TaurusCurveDialog: /home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/extra_guiqwt/plot.py:557: DeprecationWarning: TaurusCurveDialog is deprecated since 4.1. Use TaurusPlot / taurusplot launcher instead
  w = TaurusCurveDialog(edit=False, wintitle=options.window_name)

  File "/usr/lib/python3.5/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/extra_guiqwt/plot.py", line 673, in <module>
    taurusCurveDlgMain()
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/extra_guiqwt/plot.py", line 563, in taurusCurveDlgMain
    w.setModel(args)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/extra_guiqwt/plot.py", line 119, in setModel
    self.addModels(modelNames)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/extra_guiqwt/plot.py", line 165, in addModels
    linestyle=linestyle, linewidth=2)
  File "/usr/lib/python3/dist-packages/guiqwt/builder.py", line 382, in curve
    return self.pcurve(x, y, param, xaxis, yaxis)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/extra_guiqwt/builder.py", line 80, in pcurve
    curve = TaurusCurveItem(param)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/extra_guiqwt/curve.py", line 50, in __init__
    CurveItem.__init__(self, curveparam=curveparam)
  File "/usr/lib/python3/dist-packages/guiqwt/curve.py", line 332, in __init__
    super(CurveItem, self).__init__()
  File "/usr/lib/python3/dist-packages/qwt/plot_curve.py", line 185, in __init__
    self.init()
  File "/usr/lib/python3/dist-packages/qwt/plot_curve.py", line 192, in init
    self.setData(QwtPointArrayData())
  File "/usr/lib/python3/dist-packages/qwt/plot_curve.py", line 844, in setData
    super(QwtPlotCurve, self).setData(*args)
  File "/usr/lib/python3/dist-packages/qwt/plot_series.py", line 287, in setData
    self.dataChanged()
TypeError: native Qt signal is not callable
```

But I must be doing something wrong...

reszelaz:
> The code looks ok if we assume that the bytes will arrive in utf-8 encoding. I was looking for this being documented, for example here but could not find it explicitly said.

If I understand correctly, this is the `TaurusBaseWidget.getModelMimeData` who encodes using utf-8 and them PyQt just operates on bytes. If this is like this, everything looks ok.

Then to finish with the review I would need a way to test the changes from `curvesmodel.py` and `curveprops.py`. Any idea how to do it?

cpascual:
> Then to finish with the review I would need a way to test the changes from curvesmodel.py and curveprops.py. Any idea how to do it?

The changes in these to parts are not really needed (they could just be left unchanged) since they are both in deprecated code. I just changed them for coherence with the rest of taurus (and to avoid someone using them as the wrong example for other developments):

- In the case of curvesprop, it is in the (deprecated) qwt5 module which does not support python3 , so I would just test that it is not breaking anything when using py2+qt4). Furthermore, the curveprops module is a half-implemented and not used module, so I would not dedicate a second to it

- In the case of curvesmodel.py, again it is only used in deprecated code which quite certainly never got used in production, so I would not pay much attention to it

PR #893, opened 2019-03-27
Add a getValidatorFromName method in TaurusFactory

cmft:
Add a new method in TaurusFactory to get the TaurusModel validator object from a model name.
Add a helper in taurushelper to get the TaurusModel validator object  from a model name.


cmft:
If you right with these changes I will modify the changelog

cpascual:
Thanks for the addition!
It would be extra-nice if you also provided a unit test

cpascual:
Merging after adding unittest and fixing an unhandled exception

PR #895, opened 2019-03-29
Change modelsAdded signals signature

cmft:
Change TaurusModelSelectorItem.modelsAdded and
TaurusModelSelector.modelsAdded signals signature from
'QStringList' to list.

It will allow to pass XY models.


lib/taurus/qt/qtgui/panel/taurusmodelchooser.py
```diff
@@ -53,7 +53,7 @@ class TaurusModelSelector(Qt.QTabWidget):
     # TODO add action to add new TaurusModelSelectorItem
     # TODO add mechanism to allow manual plugin activation
     #      (instead of relying on installation)
-    modelsAdded = Qt.pyqtSignal('QStringList')
+    modelsAdded = Qt.pyqtSignal(list)
```
cpascual: If "QStringList" is not working, I think that it is better to use `taurus.external.qt.compat.PY_OBJECT` instead of `list`

PR #896, opened 2019-03-29
Add unit to taurus.core `__init__.py`

cpascual:
The `taurus.core.units` module is not imported in `taurus.core` and
therefore it does not appear in the API docs. Import it.

### DIFF
--- a/lib/taurus/core/init_bkcomp.py
+++ b/lib/taurus/core/init_bkcomp.py
@@ -42,6 +42,7 @@ from .taurusmanager import *
 from .taurusoperation import *
 from .tauruspollingtimer import *
 from .taurusvalidator import *
+from .units import *
 
 # enable compatibility code with tau V1 if tauv1 package is present
 try:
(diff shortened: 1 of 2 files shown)


lib/taurus/core/init_bkcomp.py
```diff
@@ -42,6 +42,7 @@
 from .taurusoperation import *
 from .tauruspollingtimer import *
 from .taurusvalidator import *
+from .units import *
```
reszelaz: It does not change this PR but in general I though that it was a good practice to avoid using plurals: unit vs. units.

cpascual: >it was a good practice to avoid using plurals: unit vs. units.

Yes, that's true, and I was aware of that when I originally implemented it. But in this case I had to decide between that and being consistent with the use that is done in pint of the word "units" (e.g. a Quantity has `.units`)

> python -c "from taurus.core import Quantity" works now

Yes, but still, for documentation purposes it is better to import from `taurus.core.units` because the other way does not work with `tauruscustomsettings.LIGHTWEIGHT_IMPORTS=True`

PR #897, opened 2019-04-01
Fix isValidName helper

cmft:
According the doc if `etypes` is passed, it returns True
only if name is valid for at least one of the given the
element types.

The method returns false when several `etypes` are given
if the first does not match.

Fix it.


cpascual:
good catch!

PR #902, opened 2019-04-09
New LED images with drawings

mbroseta:
Modified png's for the red, green, yellow, orange and blue LEDs, adding an icon inside the LED that indicates the status for color blind people.


cpascual:
Hi @mbroseta, thanks for the PR!

I see some issues with it, though:

1. your proposed changes do not allow someone to keep the "old" (icon-less) leds, and some people may complain about their existing GUI appearance being changed. So I think that it would be nice to keep the old icons and adding these as new icons with a different name, and then we should provide a mechanism for choosing (more about this later)

2. I also see an even more fundamental issue: originally the led images are general in the sense that they just define a led of a given color on and off. The meaning of that color (e.g. orange meaning warning) is an external association (which is done in e.g. `_TaurusLedControllerState.LedMap`). With your proposed change, e.g the "warning" meaning gets associated to the orange led image itself.

3. Finally, I see some images changed which do not have any icon added (e.g. the white and magenta ones). What are the modifications? If there are no practical modifications, maybe they should be omitted from the PR

So, a possible solution for 1 and 2 would be to:
- leave the existing "iconless" images as they were
- provide a new set of images with names reflecting "warning" "fault", "ok", etc. instead of colors
- create a new `_TaurusLedControllerState.LedMap` that uses the new states
- consider a mechanism allowing to use the original map instead of the new one for people who prefer to keep the original view (maybe controllable via tauruscustomsettings)

mbroseta:
Hi @cpascual ,

I think, my lats commit to this pull request hasn't been reviewed. In it, I am keeping the images256 as they are in the original taurus project and I have created a new folder for the new icons "images256icons".

At the beginning, I though to add some new functionallity to te code to let the user select between old leds style or the new ones, but that functionallityis already available by changing the LedPattern _**(setLedPatternName)**_ . So Ithat's why I have just added the new icons to a new folder. Is not enough with this?

I am totally agree that not all the icons have been modified. I have just modified the ones that I use more often.

Manolo

cpascual:
>I am keeping the images256 as they are in the original taurus project and I have created a new folder for the new icons "images256icons".

Sorry, my fault. I did  not notice the "images256icons" in the path

> that functionallityis already available by changing the LedPattern (setLedPatternName) . So Ithat's why I have just added the new icons to a new folder. Is not enough with this?

It is enough as long as we do not want to make the new pattern the default one ... so for now, yes. 

> I totally agree that not all the icons have been modified. I have just modified the ones that I use more often.

Would it make sense to add the unmodified ones as symlinks to the  original ones instead of duplicating them?
In principle git handles symlinks (even handles the case of local repos on filesystems not supporting symlinks ), but I confess that I am not familiar with whether it all works seamlessly or not...

cpascual:
It seems that we won't be putting more time into the small details left to do, so I am merging this as it is

PR #913, opened 2019-05-13
Make abstract the preAttach method

cmft:
Now preAttach is an abstract method and the taurusEvent signal
connection is moved to the __init__ method.


cmft:
Testing this PR. It does not work.

cmft:
> Testing this PR. It does not work.

Fixed. The preDetach method must to be also abstract

PR #917, opened 2019-05-24
Protect against DeviceAttribute without .value

mrosanes:
DeviceAttribute is not providing .value in some cases.
This has been seen using PyTango 9.3.0 when trying to read empty
spectrums.

Protect against DeviceAttribute not providing .value.

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -119,8 +119,10 @@ class TangoAttrValue(TaurusAttrValue):
                     for _ in range(len(shape) - 1):
                         p.value = [p.value]
 
-        rvalue = p.value
-        wvalue = p.w_value
+        # Protect against DeviceAttribute not providing .value in some cases,
+        # seen e.g. in PyTango 9.3.0
+        rvalue = getattr(p, 'value', None)
+        wvalue = getattr(p, 'w_value', None)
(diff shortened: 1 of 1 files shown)


mrosanes:
Read empty spectrum fails on taurus using pytango 9.3.0

Root problem: PyTango 9.3.0 also fails on reading empty spectrums. Most probably this is already introduced in Tango C/C++.

This failure has been found when executing Sardana spock using PyTango 9.3.0

To reproduce the problem easily without using Sardana from ipython do (in a system with PyTango 9.3.0):
<pre>
import tango
import taurus
d = tango.DeviceProxy("sys/tg_test/1")
d.short_spectrum = []
taurus.Attribute("sys/tg_test/1/short_spectrum").read()
</pre> 

This problem is related with https://github.com/tango-controls/pytango/issues/271

This PR adds a protection in Taurus when DeviceAttribute does not provide .value

cpascual:
Thanks,

Note that the fix of tango-controls/pytango#271 (in tango-controls/pytango#273) makes Tango C++ and PyTango differ in this respect, which is probably not a good permanent solution, so I would expect that the either the old (pre-9.3) behaviour is restored in Tango or that PyTango would end adapting to the new (backwards-incompatible) behaviour.

In either case, the changes in this PR would protect us.

PR #926, opened 2019-06-18
Fix #843: exception in taurusgui --new-gui with Py3

### LINKED ISSUE #843: Bug in taurusgui new-gui creation using python3
In a Conda environment, using Python 3.4.5

When executing:
taurusgui --new-gui 
and following the wizard, I get the following bug:

```console
(tau-py3) mrosanes@pc384:~/taurusguis$ taurusgui --new-gui
MainThread     INFO     2019-01-22 11:45:19,786 TaurusRootLogger: Using "PyQt4" for Qt
MainThread     INFO     2019-01-22 11:45:20,800 taurus.qt.qtgui.icon.icon: Setting Tango icon theme (from /home/mrosanes/anaconda2/envs/tau-py3/lib/python3.4/site-packages/taurus/qt/qtgui/icon/)
MainThread     INFO     2019-01-22 11:45:21,000 TaurusRootLogger: Starting execution of TaurusGui
  File "/home/mrosanes/anaconda2/envs/tau-py3/lib/python3.4/site-packages/taurus/qt/qtgui/taurusgui/appsettingswizard.py", line 1304, in initializePage
    self._xml.setText(xml)
TypeError: setText(self, object): argument 1 has unexpected type 'bytes'
```

After this, when installing the gui using 'pip install .' and trying to execute this new GUI, the Gui cannot be executed.

cpascual:
The default return value in py3 from lxml.etree.tostring() is bytes.
This caused an exception (#843) because we used it as unicode. Fix it by
explicitly requesting unicode result from tostring.
Also apply the same change in other usages of etree.tostring()


cpascual:
Apart from fixing the original bug, I added a commit fixing a related problem when loading the generated gui in py3 and also a minor metadata update.

cpascual:
IMHO, this is ready for integration

teresanunez:
I have checked it and it solves the bug.

PR #934, opened 2019-06-27
Accept unknown kwargs in taurus mixin class constructors

cpascual:
BaseConfigurableClass, TaurusBaseComponent, TaurusBaseWidget and TaurusBaseWritableWidget are used as mixin classes, mostly with QtObject-derived classes.

In order to support cooperative multiple inheritance with the super() statement, they need to be able to accept unknown kwargs in their constructors. 
See: https://www.riverbankcomputing.com/static/Docs/PyQt5/multiinheritance.html

This allows the following pattern:

```python
class Foo(QObject, TaurusBaseComponent):
    def __init__(self, bar=None, **kwargs):
        super(Foo, self).__init__(**kwargs)  # <- py2+py3
        # super().__init__(**kwargs)  # <- py3 only

        self.bar = bar
```

### DIFF
--- a/lib/taurus/qt/qtcore/configuration/configuration.py
+++ b/lib/taurus/qt/qtcore/configuration/configuration.py
@@ -137,7 +137,7 @@ class BaseConfigurableClass(object):
     # the latest element of this list is considered the current version
     _supportedConfigVersions = ("__UNVERSIONED__",)
 
-    def __init__(self):
+    def __init__(self, **kwargs):
         self.resetConfigurableItems()
 
     @staticmethod
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -110,7 +110,7 @@ class TaurusBaseComponent(TaurusListener, BaseConfigurableClass):
 
     taurusEvent = baseSignal('taurusEvent', object, object, object)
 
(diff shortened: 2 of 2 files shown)


cmft:
I reviewed the code. It seems ok for me

stanislaw55:
Hi,
AFAICS, best practises and my experience says that Taurus derived widgets should call parent's `__init__` method directly, not by `super`. 

Does this change make it possible to call parent's `__init__` method with `super` safely?

cpascual:
> AFAICS, best practises and my experience says that Taurus derived widgets should call parent's __init__ method directly, not by super.


You are right in that we always recommended calling the parent constructors explicitly in the case of "taurifying" a QtWidget. We have kept this criterion even when PyQt  recommends using super().

> Does this change make it possible to call parent's __init__ method with super safely?

The change makes it possible, but at this moment I do not feel confident enough to guarantee safety or to revert our general recommendation and favour super instead. However,  I've personally identified some particular situations (mostly working with PyQtGraph) in which using super() was more convenient and they are the reason for proposing this PR: to open the door to experiment with super in some specific cases (but without making it the recommended general solution yet).

stanislaw55:
Thanks for explanation.
Great to hear that experimentation is possible now.

tiagocoutinho:
LGTM

PR #940, opened 2019-07-02
Extra config options

stanislaw55:
Summary of changes:
 - add possibility to specify icon in `PanelDescription` (was possible only trough `createPanel` only)
 - add possibility to specify `modelInConfig`, `modifiableByUser` and `formatter` in both `PanelDescription` and `createPanel` method
 - add possibility to pass extra configuration options in form of keyword arguments to `PanelDescription` constructor. It widget has attribute named like passed argument name, its value is set to value passed by keyword argument value.


cpascual:
Hi @stanislaw55 , thanks for the contribution!

In order to understand the purpose of the PR, can you please give some context? (e.g. what use do you have in mind for the new arguments?

stanislaw55:
Hi,
those already present Taurus configuration options (`icon`, `modelInConfig`, `modifiableByUser` and `formatter`) I found missing in configuration file (.py or .xml). Honestly, I do not know if this was desing decision or not but in few applications it was useful to have them configured. The workaround was to use `defaultFactorySettingsFile` (i. e. `default.ini`) but in my opinion it looks very put configuration of different options in two different places.

About extra configuration options: it is fragile concept but powerful one. In many applications, I found myself following this approach: create one abstract/general widget with all logic required but missing some configuration steering this logic. Then, create few specialized widgets inheriting from the abstract/general one with no logic, just static variables filling role of configuration. This was quite common technique applied by me but it seems inappropriate (mixing configuration and logic in source code). But giving possibility to provide extra options in `PanelDescription`, I would avoid this pattern. 

Also, I believe that this changes will be helpful for other developers using Taurus as this extends possibilities of configuration. 

Another thing is that those changes are extension, so not impacting others not using these features.

The last but not the least, this PR is IMO ready for merge from *technical* point of view but I'm not sure if the *approach* taken to implement those changes is correct. Looking for feedback about that.

cpascual:
Thanks @stanislaw55 for the explanation.

IMHO the changes are justified. I'll come back with comments about implementation when I get some time to review (right now I am busy with sardana and the Jul19 release of Taurus). Speaking of which... is it important for you to get this PR included in the Jul19 release? or is it ok if we integrate it after the release? (we are a bit tight on time)

stanislaw55:
Hi,
no need to rush with merging, better late then buggy, as one would say.

cpascual:
Hi, I finally managed to test this and review the implementation.

As I said, I think that the objectives of this PR are justified, but I am not too sure about some implementation details. I'll write my concerns as comments to the code

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
```diff
@@ -665,6 +669,11 @@ def createPanel(self, widget, name, floating=False, registerconfig=True, custom=
             self.info('Panel with name "%s" already exists. Reusing.' % name)
             return self.__panels[name]
 
+        if isinstance(widget, TaurusBaseComponent):
+            widget.setModifiableByUser(modifiablebyuser)
```
cpascual: Even if we disregard my previous comment about these options not belonging to `createPanel`, note that this implementation would break things in the case of widgets that are implemented with customized values of the affected properties (e.g., the `TaurusTrend2DDialog` is implemented with `modifiableByUser=True` by default, but with this implementation it would be set to False). 
IMHO, `createPanel()` should respect the widget's defaults at least unless explicitly overriden.

lib/taurus/qt/qtgui/taurusgui/utils.py
```diff
@@ -397,11 +413,19 @@ def fromPanel(panel):
         else:
             # ignore other "model" attributes (they are not from Taurus)
             model = None
+        icon = panel.icon
+        model_in_config = panel.model_in_config
+        modifiable_by_user = panel.modifiablebyuser
+        formatter = panel.formatter
```
cpascual: I think that the `formatter` that makes sense to store is `panel.widget.formatter`, not `panel.formatter`

Also note that the members of the panel description should be serializable in order to be able to store them as xml. In the case of the formatter, it could be a string, but it could also be a *function*. In the latter case, some conversion would be needed to serialize it as a string.

lib/taurus/qt/qtgui/taurusgui/utils.py
```diff
@@ -365,7 +365,23 @@ class PanelDescription(TaurusGuiComponentDescription):
...
+
+        Additionally, extra configuration options can be passed in constructor as key word arguments.
+        Proper widget attributes will be set to corresponding values
+        """
         self.instrumentkey = kwargs.pop('instrumentkey', None)
+        self.icon = kwargs.pop("icon", None)
+        self.model_in_config = kwargs.pop("modelinconfig", False)
+        self.modifiable_by_user = kwargs.pop("modifiablebyuser", False)
+        self.formatter = kwargs.pop("formatter", 'taurus.qt.qtgui.base.taurusbase.defaultFormatter')
+        self.extra_config = kwargs
```
cpascual: In this proposed implementation, `extra_config` (which refers to widget attributes/properties to be set) is obtained implicitly as whatever kwargs not explicitly consumed in the `__init__`. In my experience it is better to require the caller to be explicit (e.g., by calling the constructor as `PanelDescriptor(icon=..., widget_properties={"foo"=1,"bar":2})`.

While the implicit version is more concise, it also becomes a PITA to maintain (think of refactorings, deprecations, etc.)

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
stanislaw55: Indeed, default values should be respected if no overriding values are provided in PanelDescription.

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
```diff
@@ -629,7 +629,8 @@ def removePanel(self, name=None):
         self.debug('Panel "%s" removed' % name)
 
     def createPanel(self, widget, name, floating=False, registerconfig=True, custom=False,
-                    permanent=False, icon=None, instrumentkey=None):
+                    permanent=False, icon=None, instrumentkey=None, modelinconfig=False,
```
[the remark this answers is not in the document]
stanislaw55: I mixed up panel from TaurusGui and PanelDescription too much.
I'll change to storing those settings only in PanelDescription and applying in `TaurusGui._loadCustomPanels`

lib/taurus/qt/qtgui/taurusgui/utils.py
stanislaw55: I trust your experience. Will change.

I think `widget_properties` is better (more descriptive) name for these options

lib/taurus/qt/qtgui/taurusgui/utils.py
stanislaw55: Actually name `formatter` is not the best one for what I wanted to achieve. More descriptive would be `widget_formatter`. I've accidentally use wrong attribute of `PanelDescription` but now name is changed so name conflict is resolved.

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
```diff
@@ -833,9 +833,6 @@ def createCustomPanel(self, paneldesc=None):
             w.setCustomWidgetMap(self.getCustomWidgetMap())
         if paneldesc.model is not None:
             w.setModel(paneldesc.model)
-        if isinstance(w, TaurusBaseComponent):
```
cpascual: Removing this would break applications that rely on the user to be able to modify the custom widgets after creation. For example, consider the following pretty common use case: a user clicks on "Panel-->New Panel --> selects a TaurusForm and gives it a name -->accepts the dialog --> attempts to drag and drop values from other forms.

cpascual: I am preparing some modifications to the code and I'll take care of this, so do not act on this for now.

cpascual:
Hi, I did some more review and modified the code myself with some suggestions, but it seems that I cannot push to the PR of this branch directly, so instead I proposed the changes as a PR to *your PR*:
https://github.com/synchrotron-solaris/taurus/pull/3 
please review and merge it so that we can continue with the integration here

cpascual:
Travis builds fail for py2qt4, but it is an unrelated SIGSEV issue. So I am merging.

Thanks @stanislaw55 for your work and your patience!

PR #941, opened 2019-07-04
Document missing modules 

### LINKED ISSUE #898: [doc] some classes are not being documented in the API
I miss the docs for `taurus.core.tango.TangoAttributeNameValidator` and other validators,
but there are likely more missing

cpascual:
Fix #898:

ModuleExplorer.exploreModule discovers subpackages but fails to explore submodules implemented as single files. This causes many classes to be undocumented unless they are explicitly imported in some parent module.

Fix this by discovering single-file submodules (but filtering out those which *are* already imported since those are not considered part of the API


tiagocoutinho:
CI seems to fail only because of  pint version issue (solved in #956)

PR #946, opened 2019-07-16
Relax TaurusDevPanel state checking

### LINKED ISSUE #882: Wrong DevState management in TaurusDevPanel
<s>TaurusDevicePanel</s> TaurusDevPanel manages TaurusDevState as if it was related to Tango communication

TangoDevice instead is assigning TaurusDevState depending of Hardware state (independently if it communicates or not).

This discrepancy does not allow to display devices requiring user action (e.g. a Disabled device that needs to be re-enabled via Tango command).

cmft:
TaurusDevPanel is not able to show attributes and command panels
of given device if it is not in `Ready` state.

This is not coherence with other tools such Jive and Taurus form.

Show these panels in TaurusDevPanel unless the state is `Undefined`.

Fix #882

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusdevicepanel.py
+++ b/lib/taurus/qt/qtgui/panel/taurusdevicepanel.py
@@ -589,7 +589,7 @@ class TaurusDevPanel(TaurusGui):
         dev = self.getModelObj()
         state = dev.state
         # test the connection
-        if state == TaurusDevState.Ready:
+        if state != TaurusDevState.Undefined:
             msg = 'Connected to "%s"' % devname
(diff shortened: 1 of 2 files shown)


cpascual:
Hi @cmft , thanks for the PR.

However, I am not sure about this approach. It might work but, as I actually commented in https://github.com/taurus-org/taurus/issues/882#issuecomment-473869744 , I think that we should not check the state at all (basically, why should we treat `UNDEFINED` differently than e.g. `FAULT`), but instead simply try to show the attributes and handle the failure if it fails.

cpascual:
LGTM

PR #949, opened 2019-07-18
Fix cyclic import when importing TaurusGui

### LINKED ISSUE #947: Py3 Qt5 TaurusGui import problem
Bug in Taurus when using Python3 and Qt5, and trying to import TaurusGui
This bug exists on Linux and on Windows. 

Note: It is shadowed if taurus_pyqtgraph is installed, because the chain of recursive imports is broken then. If taurus_pyqtgraph is not installed, in a clean conda environment, the bug is reproduced. It can be reproduced with the recipe "Create a CONDA environment (with py3 and Qt5)":
https://github.com/taurus-org/taurus/wiki/Recipes

from taurus.qt.qtgui.taurusgui import TaurusGui

ImportError: cannot import name 'TaurusGui'

This bug is affecting the launching of newly created taurus GUIs.

mrosanes:
Create helper functions in order to fix cyclic import appearing
when importing:
from taurus.qt.qtgui.taurusgui import TaurusGui

This problem affects the launching of newly created Taurus Guis.

This commit fixes it.

Fixes #947

### DIFF
--- a/lib/taurus/qt/qtgui/taurusgui/paneldescriptionwizard.py
+++ b/lib/taurus/qt/qtgui/taurusgui/paneldescriptionwizard.py
@@ -35,7 +35,6 @@ import weakref
 from taurus.qt.qtgui.taurusgui.utils import PanelDescription
 from taurus.qt.qtgui.icon import getCachedPixmap
 from taurus.qt.qtgui.input import GraphicalChoiceWidget
-from taurus.qt.qtgui.panel import TaurusModelChooser
 from taurus.qt.qtgui.base import TaurusBaseComponent, TaurusBaseWidget
 from taurus.qt.qtcore.communication import SharedDataManager
 from taurus.qt.qtcore.mimetypes import TAURUS_MODEL_LIST_MIME_TYPE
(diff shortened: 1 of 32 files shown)


mrosanes:
This PR fixes the issue https://github.com/taurus-org/taurus/issues/947

lib/taurus/qt/qtgui/taurusgui/paneldescriptionwizard.py
```diff
@@ -48,6 +47,11 @@
 __all__ = ["PanelDescriptionWizard"]
 
 
+def modelChooserDlg(*args, **kwargs):
```
cpascual: If possible, I would prefer to avoid adding public functions at the module level.

In this case I think that we can simply remove this method and add the import at the beginning of `showModelChooser()`

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
```diff
@@ -58,6 +56,16 @@
 __docformat__ = 'restructuredtext'
 
 
+def QDoubleListDlg(*args, **kwargs):
```
cpascual: Same as in my previous comment, but in this case the solution is not as simple.

Still, I would prefer adding import lines where ExternalAppEditor or QDoubleListDlg are used.

Alternatively I would rename these functions to add a '_' prefix (and of course rename their usages accordingly).

Also, in this last case, I think that the functions should be clearly documented as "workarounds", and a TODO tag added to try to fix this by a refactoring of the imports.

PR #951, opened 2019-07-18
Use encoding in etree.tostring

cmft:
Taurus expects string when it calls etree.tostring method.
In Python3 this method returns bytes instead string if the
encoding argument is not given.


cpascual:
Updated with latest develop changes to avoid doc issues in travis

PR #954, opened 2019-07-18
Fix imports of QtWebkit

cpascual:
taurus.external.qt.QtWebkit fails to import due to wrong module name.
Fix it.


tiagocoutinho:
CI seems to fail only because of sphinx doc (and this is solved by #941) and pint version issue (solved in #956)

PR #955, opened 2019-07-19
Fix launch taurus designer on windows (closes #871)

### LINKED ISSUE #871: taurusdesigner does not launch in windows10 with py3 and qt5
launching `taurusdesigner on a windows 10 with py3 qt5 yields 
ModuleNotFound: no module named 'PyQt4'

Note that this does not happen with debian py3 qt5

tiagocoutinho:
* Do a best effort on windows to find Qt designer
* Remove explicit import of PyQt4


cpascual:
I fixed an indentation problem that made the docs fail to test and it LGTM

PR #956, opened 2019-07-19
Improve TaurusLabelTest2 to support various pint versions

cpascual:
pint v0.8 fixed a bug which had, as side efect a change in str
conversion which breaks TaurusLabelTest2. Adapt the expected value
of a test to handle different pint versions.


tiagocoutinho:
CI seems to fail only because of sphinx doc (and this is solved by #941).

PR #961, opened 2019-07-23
Custom widget map preserved

stanislaw55:
Summary of changes:
 - do not override custom widget map of widget in TaurusGui. Update default map with widget's map instead
 - update custom widget map only in one place (`_loadCustomPanels`)


cpascual:
I like the idea of *updating* instead of *overriding* the widget maps of the widgets being added to panels in the guis, but I am not sure that we can do this only in `_loadCustomPanels`.

For example: note that in the original implementation, the map is set in `DockWidgetPanel.setWidgetFromClassName()`, which is called by `DockWidgetPanel.applyConfig()` prior to applying the config. I am afraid that If the map is not set (or updated), the config may fail to apply because the widgets may differ from those for which the config was saved.

stanislaw55:
Hi @cpascual 
First of all, thanks for reviewing!
You might be right about that. However, I think it would be nice to have in place in code but I do not see any way to achieve this.

Notwithstanding, I've also added updating custom widget map in `DockWidgetPanel.setWidgetFromClassName`

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
```diff
@@ -194,6 +191,10 @@ def applyConfig(self, configdict, depth=-1):
                 'widgetClassName'), modulename=configdict.get('widgetModuleName', None))
             if isinstance(self.widget(), BaseConfigurableClass):
                 self.widget().applyConfig(configdict['widget'])
+            if hasattr(w, "setCustomWidgetMap") and hasattr(w, "getCustomWidgetMap"):
```
cpascual: Is it me, or `w`  is not defined here?  (I think it should be `self.widget()`)

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
```diff
@@ -194,6 +191,10 @@ def applyConfig(self, configdict, depth=-1):
...
+                tmp = self._mainwindow.getCustomWidgetMap()
```
cpascual: a very minor/unimportant comment: it would be better to use a more descriptive name for the `tmp`  variable (maybe `widget_map`

lib/taurus/qt/qtgui/taurusgui/taurusgui.py
```diff
@@ -832,8 +833,6 @@ def createCustomPanel(self, paneldesc=None):
             if not ok:
                 return
         w = paneldesc.getWidget(sdm=Qt.qApp.SDM, setModel=False)
-        if hasattr(w, 'setCustomWidgetMap'):
-            w.setCustomWidgetMap(self.getCustomWidgetMap())
```
cpascual: shouldn't we also update the map in this case?

I think a similar code to that of `applyConfig` (i.e., setting a map that is obtained from updating the widget's with the windows' ) should also be done here instead of removing it. ... or am I missing something?

cpascual:
> Therefore I propose to close this PR without merging.

In order to clean-up issues for Jul20 release, I am closing this. Feel free to reopen if you consider necessary

PR #988, opened 2019-08-06
Create cli subcommand "taurus check-deps"

### LINKED ISSUE #966: create cli subcommand "taurus check-deps"
Right now the status of taurus dependencies can be checked with:

`python -c 'import taurus; taurus.check_dependencies()'`

It would be nice to implement a click subcommand to execute the same function

cmft:
Right now the status of taurus dependencies can be
checked with:

python -c 'import taurus; taurus.check_dependencies()'

Add new click subcommand to execute the same function.

Fix #966


cpascual:
Thanks @cmt !
One note: we also need to modify the getting-started docs where we refer to dependence checking

PR #991, opened 2019-08-23
Expose some API from taurus.cli

cpascual:
Expose register_subcommands() and taurus_cmd() in taurus.cli

This is useful to add cli integration tests in taurus plugins that provide subcommands


cmft:
Merged thanks

Issue #851, opened 2019-01-29
[Travis] Travis tests are not robust


cpascual:
The travis runs of our testsuite fail without known reason too often... and they eventually work when re-run.
This is not new (see #261, #497, #601) and we have been doing workarounds or just enduring it for too long. Currently the majority of the Travis builds have at least one job failing without proper reason and it may require up to a dozen manual reruns until it goes green.

The typical errors seen are:

1. just hangs in the middle of the testsuite. Typically around the same lines, but not always. It seems a time-related abort more than related to some specific test
2. tango communication errors (these look like race conditions)
3. poor isolation between tests 

Class 1 errors seem Travis-specific. 
Class 2 errors are definitely more frequent when run on Travis than in our local machines
Class 3 errors can be seen also locally. Maybe running the test in isolated processes (as in `pytest --boxed`) could help

I am opening this issue to be able to refer to it from other PRs and also to ask for help in fixing this once and for all.

cpascual:
Since we identified that most of the spurious failures came from the cpascual/taurus-test:debian-stretch-py3 image, we just flagged the tests on that image as "allowed to fail" ( see e26c00659)

That image tests the py3+qt4 combination on stretch . 
The py2+qt4 and py3+qt5 combinations seem to be much more robust and are still enforced.

TODO: investigate the possibility that this is a problem with the docker image itself.
TODO: consider updating the buster image and enforcing it (now that buster goes into freeze)

cpascual:
After the merge of #994,  it seems that most sources of instability are gone (or can be worked around using pytest's flaky flags).

I propose to keep an eye during a few integration rounds and close this if this impression is confirmed

PR #994, opened 2019-08-29
Refactor the test CI: use tox, pytest and tango-cs-docker

cpascual:
This PR proposes a complete refactoring of our CI in travis.

Improvements:
1. Based in [tox](https://tox.readthedocs.io) + [tox-conda](https://github.com/tox-dev/tox-conda) (tests can be reproduced locally in a similar way as they are executed by travis). 
2. The tango support in travis is provided by [tango-cs-docker](https://github.com/tango-controls/tango-cs-docker) *running as an independent service*. We do not use or need the taurus-test images now.
3. Facilitates adding new tests and test environments: this PR already adds tests for various interesting combinations of versions of python (2.7, 3.5, 3.6, 3.7), and of Qt bindings (PyQt4, PyQt5 and Pyside2), as well as flake8 tests and doc builds.
4. Tests are run with [PyTest](https://pytest.org/) instead of unittest (the tests are not changed for now, just run with PyTest even if they are implemented with unittest). In practice, we see that the travis builds seem more robust now(maybe #851 is fixed by this?). It also opens the door to using many interesting features of PyTest (fixtures, parameterizations, pytest-qt,...)
5. We can use tox to deploy local environments for test and development:
   ```
   tox --devenv -e py37-qt5 taurusenv
   ```
6. A flake8 test is implemented, but for now it is allowed to fail (I intend to submit another PR fixing all flake8 issues in taurus) 
 
Notes for integrators:
- The results can be seen in https://travis-ci.org/taurus-org/taurus/builds/578400811
- The docs generated by travis are in https://taurus-org.github.io/taurus-doc/v-pytest
- One can also run all the tests locally (on a system with conda available) by running `tox` (this takes long, so one can list the environments with `tox -l` and run only a selectionwith `tox -e`)


mrosanes:
Great work!

cpascual:
Hi @taurus-org/integrators , can someone review an integrate this ? 
It would be nice to have it integrated to facilitate the cleaning flake8 issues

teresanunez:
Hi carlos, I have had a look and I agree with the changes you mention and the outputs are
 fine to me. But I can not tested. if this is enough I could integrate it, but perhaps somebody else
 wants to test it first. Many thanks for the work.

cpascual:
Hi @teresanunez, thanks for reviewing.

In principle the tests are done already by travis, so it is more about agreeing in the objective and implementation.

In any case, running the tests locally, should just be a matter of:
- having (mini)conda installed
- creating a conda env with `tox-conda` in it: 
```
conda create -n toxconda tox-conda
```
- From the toxconda environment, run all the tox targets of taurus (or select some with -e):

```
conda activate toxconda
cd <taurusdir>  # taurusdir is a checkout of this PR
export TOX_TESTENV_PASSENV=TANGO_HOST  # you can skip this if you define the tango host in a /etc/tangorc file
tox   # this will run all the targets. use -e to select only a few
```

cpascual:
The last changes from @cmft are a good addition (they help running tox locally)

reszelaz:
Both @cmft and myself had run the test locally and no issues were observed. I do not have time to review the changes carefully but since there is already agreemend from @teresanunez I simply merge it now.

PR #1010, opened 2019-09-26
Improve wheel edit handling of formats and digits

cpascual:
This PR tackles 3 different issues related to the wheel edit widget:
- usage of tango-centric API
- poor synchronization between the wheel edit and its internal editor
- user cannot control (via GUI) the number of digits shown by the widget

## usage of tango-centric API

For example the following code: will raise an exception and deprecation warnings:

```python
from taurus.qt.qtgui.input import TaurusWheelEdit

if __name__ == '__main__':
    import sys
    w = TaurusForm()
    w.setModel( ['sys/tg_test/1/ampli'])
    w[0].setWriteWidgetClass(TaurusWheelEdit)
    w.show()
    sys.exit(app.exec_())
```

Raises an exception:
```ValueError: '%6.2e' format unsupported```
And also a deprecation warning:
 ```DeprecationWarning: format is deprecated since 4.0.4. Use format_spec or precision instead```

This PR fixes it, **but in doing so, it removes the feature that the format is used to initialize the number of digits** (this is mitigated by the implementation of user-access to setting the number of digits, see below)

## poor synchronization between the wheel edit and its internal editor

The internal editor widget allows to type in values that cannot be represented by the wheel widget. When this occurs and return is pressed, the value is silently discarded, which can be confusing.

This is fixed by making the internal editor enforce the same limitations as the wheel when editing.

## user cannot control (via GUI) the number of digits shown by the widget

This PR adds an action (accessible via context menu of the wheel edit widget) that allows the user to set the digit count.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ develop branch) won't be reflected in this file.
 ### Changed
 ### Deprecated
 ### Fixed
+- Several issues in TaurusWheelEdit (#1010)
 
 ## [4.6.1] - 2019-08-19
 Hotfix for auto-deployment in PyPI with Travis. No other difference from 4.6.0.
--- a/lib/taurus/qt/qtgui/input/qwheel.py
+++ b/lib/taurus/qt/qtgui/input/qwheel.py
@@ -220,6 +220,10 @@ class QWheelEdit(Qt.QFrame):
         self._setValue(0)
 
         self._build()
+        self.setDigitCountAction = Qt.QAction("Change digits", self)
+        self.setDigitCountAction.triggered.connect(self._onSetDigitCount)
+        self.addAction(self.setDigitCountAction)
+        self.setContextMenuPolicy(Qt.Qt.ActionsContextMenu)
 
     def _getMinPossibleValue(self):
         """_getMinPossibleValue(self) -> None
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by cmft** [comment]
Testing with python2 and qt4 I fond issues:
 
*  when you try to set a higher value than positive limit using the upper arrows. The reported message is
`cannot be represented in current widget.  Tip: try changing the number of digits from context menu`  when it should be related to cross the limit (the same problem happens with negative limit).
image

* Also a message is reported when you press a down arrow in position that not make sense (when the value is 0); these click should be ignored but a message is reported.  

image

**Comment by cmft** [comment]
There is another issue, If you set this value `+088888888.124` in a model without limits and then you click in the last down arrow (in the left side) the write value is set to `-011111111.876`

image

**Comment by cpascual** [comment]
I agree with the issues reported about the message with the error message. I'll try to improve that.

Regarding the [other issue](https://github.com/taurus-org/taurus/pull/1010#issuecomment-535856188), I agree in that it may be surprising, but it is in fact the "intended" behaviour of the widget: the arrows blindly add/substract `10**n` where `n` is their position counting left from the decimal point (e.g.: if the value is 88 and you click on the 3rd down-arrow, you'll substract 100, so you end up with -12).

An alternative behaviour would be to ignore down-arrow clicks if the digit is 0 or up-arrow clicks if the digit is 9... This would be more intuitive (and I would not mind it), but it would make it more difficult to us the wheel for "manual scans" (which is one of the use cases for this widget)

**Comment by cmft** [comment]
> An alternative behaviour would be to ignore down-arrow clicks if the digit is 0 or up-arrow clicks if the digit is 9... This would be more intuitive (and I would not mind it), 

Agree

> but it would make it more difficult to us the wheel for "manual scans" (which is one of the use cases for this widget)

IMO, a miss click in down-arrow during a "manual scans" can be dangerous

**Comment by cpascual** [comment]
> IMO, a miss click in down-arrow during a "manual scans" can be dangerous

I don't agree: if going to negative values is dangerous, this should be protected by the attribute range.

**Comment by tiagocoutinho** [comment]
I run the following example:
```python

from taurus.qt.qtgui.input import TaurusWheelEdit

app = Qt.QApplication([])
p = Qt.QWidget()
l = Qt.QHBoxLayout(p)
w = TaurusWheelEdit()
w.setModel('sys/tg_test/1/double_scalar')
ll = TaurusLabel()
ll.setModel('sys/tg_test/1/double_scalar')
l.addWidget(w)
l.addWidget(ll)
p.show()
app.exec_()
```
and both the validator and the precision showed don't seem to work.

However, it works when I start a taurus form!

tiagocoutinho:
Last update: it seems the above code also doesn't work on the develop branch.

Maybe we can create another issue for this.

cmft:
TaurusWheelEdit does not work cause the first config event seems to be lost. After forcing one, it works at expected.

cpascual:
> Maybe we can create another issue for this

Since it is an existing issue, I agree in not blocking this one

tiagocoutinho:
Ok, merging then.

PR #1012, opened 2019-09-30
Do not force Qt theme on linux by default

cpascual:
The value of `QT_THEME_FORCE_ON_LINUX` is set to `True` by default in
`tauruscustomsettings.py`, which is inconsistent with what is documented
in that same module. Set it to `False` by default.

### DIFF
--- a/lib/taurus/tauruscustomsettings.py
+++ b/lib/taurus/tauruscustomsettings.py
@@ -134,7 +134,7 @@ QT_THEME_NAME = 'Tango'
 
 #: In Linux the QT_THEME_NAME is not applied (to respect the system theme)
 #: setting QT_THEME_FORCE_ON_LINUX=True overrides this.
-QT_THEME_FORCE_ON_LINUX = True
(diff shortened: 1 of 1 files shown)


reszelaz:
I have tested it quite a lot on the Debian Buster using this branch while trying ICALEPCS training materials (50 % done). And everything looks pretty ok. Some minor comments:

I see that we lost the "Add to favourites" icon in the macroexecutor.

Also I see that one of the icon, the one which is showed in the KDE menu -> Applications -> System appears in many places, at least:
* In TaurusDbTreeView widget:  "By device", "Alias", "Device"
* In the Taurus GUI Panels -> Switch Temporary/Permanent Panel...

reszelaz:
I've completed all tests. With this change https://github.com/sardana-org/sardana/pull/1208 the issue of the favourites icon get solved.
Regarding the TaurusDbTreeView widget, after talking with @cpascual, we could avoid using the theme icon and fix a more meaningful icon.
I'm merging.

PR #1024, opened 2019-11-07
Fix TaurusImageDialog.setModel Qt slot in py2

### LINKED ISSUE #1023: slot setModel from TaurusImageDIalog is not exposed to qtdesigner
Hi,

I try to upgrade a 5 years old application:
- from Python 2.6 to Python 2.7
- from Qt4 to Qt5
- from Taurus 3.4 to Taurus 4.5.1

I use to use a lot qt designer, and signals.

Unfortunately it looks like things are broken, but i can't really explain why.

My script are rejected when it tries to connect a string signal to `setModel` of `TaurusImageDialog`. I use python scripts with qtdesigner files to describe signal connections.

`setModel` looks to be correctly tagged.
```
    @Qt.pyqtSlot(str)
    def setModel(self, model):
```

But in Qt designer (`taurusdesigner`) the `setModel` slot is not exposed.

taurus-taurusimagedialog-setmodel

Any idea why?
Thanks a lot.

(yet i only try Taurus 4.5.1 version, but i guess i could use the last release without problem)

cpascual:
The setModel slot of TaurusImageDialog is not exposed in py2 due to
a wrong py2-py3 compatibility definition. Use `"QString" ` instead of
`str` for defining the slot in a py2+3 compatible way.

Fixes #1023

### DIFF
--- a/lib/taurus/qt/qtgui/extra_guiqwt/plot.py
+++ b/lib/taurus/qt/qtgui/extra_guiqwt/plot.py
@@ -474,7 +474,7 @@ class TaurusImageDialog(ImageDialog, TaurusBaseWidget):
         '''reimplemented from :class:`TaurusBaseWidget`'''
         return taurus.core.taurusattribute.TaurusAttribute
 
-    @Qt.pyqtSlot(str)
+    @Qt.pyqtSlot("QString")
     def setModel(self, model):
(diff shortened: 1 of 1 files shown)


cpascual:
@taurus-org/integrators: This is ready for integration

cmft:
Good catch!! Merged. Thanks!!

PR #1032, opened 2019-11-18
Fix TaurusValueComboBox does not show set value

cmft:
When a TaurusValueComboBox widget is initialized it does not show
the set value.

The setValue method does not find the given value <s>since it saves
internally as string. Fix it doing a cast.</s>

Update by @cpascual : It is more likely related to [this](https://stackoverflow.com/questions/12837701/pyqt-how-do-i-find-tuples-using-qcomboboxfinddata)

### DIFF
--- a/lib/taurus/qt/qtgui/input/tauruscombobox.py
+++ b/lib/taurus/qt/qtgui/input/tauruscombobox.py
@@ -115,7 +115,7 @@ class TaurusValueComboBox(Qt.QComboBox, TaurusBaseWritableWidget):
         Set the value for the widget to display, not the value of the
         attribute.
         """
-        index = self.findData(value)
+        index = self.findData(str(value))
         self._setCurrentIndex(index)
 
     def updateStyle(self):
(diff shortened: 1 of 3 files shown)


cpascual:
Just for future reference: a script to reproduce the problem is:

```python
from taurus.qt.qtgui.input import TaurusValueComboBox
import taurus
import sys
w = TaurusValueComboBox()
model = 'sys/tg_test/1/short_scalar_w'
a = taurus.Attribute(model)
a.write(2)
names = [('name0', 0), ('name1', 1), ('name2', 2), ('name3', 3)]
w.addValueNames(names)
w.setModel(model)
w.show()
sys.exit(app.exec_())
```

cpascual:
I [added a unit test](https://github.com/taurus-org/taurus/pull/1032/commits/b0c32106d14cb6ef52d80ee62ae7a7e1f0303529) that tests some cases more complex than those in the snippet above, and it shows that just casting to a str [is **not** enough](https://travis-ci.org/taurus-org/taurus/builds/614117470). 
In 6736f01 I propose a different approach that should pass the test. @cmft , can you please review it?

cpascual:
Hi, I just noticed that taurusform does not behave well with this in python2. (it works well in py3)

I'll try to fix it before merging

cmft:
I agree with the changes and seems that everything work. I will merge it

PR #1033, opened 2019-11-20
Register degreeC (= °C = ºC = degC =celsius) in the units registry

cpascual:
Make celsius display as "°C" in taurus. Do it by registering a new
unit (degreeC) equivalent to degC in `taurus.core.units.UR`
This is a workaround until pint properly supports this. See:
https://github.com/hgrecco/pint/issues/546


cpascual:
@sergirubio : you are probably interested in this.

Note, until this is merged, you can simply do the following workaround in your code:

```python
from taurus.core.units import UR
UR.define(u'degreeC = kelvin; offset: 273.15  = °C = ºC = degC = celsius')
```

MikeFalowski:
> @sergirubio : you are probably interested in this.
> 
> Note, until this is merged, you can simply do the following workaround in your code:
> 
> ```python
> from taurus.core.units import UR
> UR.define(u'degreeC = kelvin; offset: 273.15  = °C = ºC = degC = celsius')
> ```

Thanks @cpascual, I was just looking for this, you saved me a lot of time.

cpascual:
Hi, I just noticed that TaurusForm does not behave well with unicode characters in the units in python2. (it works well in py3)


b17df86 should fix it

cpascual:
Tip: when defining the unit for an attribute in Tango (e.g. via jive) some clients may not deal well with unicode. In such case, use "celsius" or "degC" as the unit and Taurus will display it as `°C`

sergirubio:
@MikeFalowski , @cpascual , Many thanks for the help!

PR #1038, opened 2019-12-10
Support spyder v4

cpascual:
The taurus.qt.qtgui.editor module fails to import with spyder 4.
Adapt its imports to support spyder 3 and 4 simultaneously.


cmft:
Merged. Thanks

PR #1039, opened 2019-12-10
Improve set formatter dialog

### LINKED ISSUE #1035: Improve set formatter dialog for taurus forms
The `TaurusForm` allows the user to change the formatter (per-item and to all items), but the dialog is very crude (just a text input), which is not user friendly.

It would be nice to improve this dialog with, at least the following:

- add a user-friendly tooltip explaining usage
- make it an editable combobox with pre-set options such as `{:2.3e}`, `{:.5f}`, `TangoFormatter`, ,...

Note: the formatter classes to be offered should be gathered from a list of "registered formatters"  (i.e. an entry point (e.g. `taurus.qt.formatters`) should be declared and plugins should register their formatters into it.

cmft:
* This PR adds a new entry point for registering Taurus formatters.
* Improves the usability of taurus formatter dialog (using the entry points)

Fixes #1035

### DIFF
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -1384,8 +1384,12 @@ class TaurusBaseWidget(TaurusBaseComponent):
         except ValueError:
             ind = 0
 
+        uri = "http://taurus-scada.org/devel/api/taurus/qt/qtgui/base/_TaurusBaseComponent.html"
+        help = "<a href=\"{}\"> Find more info in doc</a>".format(uri)
+
+        msg = "Choose/Enter a formatter: {}".format(help)
         formatter, ok = Qt.QInputDialog.getItem(self, "Set formatter",
-                                                "Choose/Enter a formatter:",
+                                                msg,
                                                 list(self._formatters.keys()),
                                                 current=ind,
                                                 editable=True)
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -29,6 +29,7 @@ should inherit to be considered valid taurus widgets."""
 
 import sys
 import threading
+import pkg_resources
 from types import MethodType
 from future.builtins import str
 from future.utils import string_types
@@ -37,6 +38,7 @@ from taurus.external.qt import Qt
 from enum import Enum
 
 import taurus
+from taurus import warning
 from taurus.core.util import eventfilters
 from taurus.core.util.timer import Timer
 from taurus.core.taurusbasetypes import TaurusElementType, TaurusEventType
@@ -64,6 +66,23 @@ __docformat__ = 'restructuredtext'
 
 
 DefaultNoneValue = "-----"
+
+
+
+# ------------------------------------------------------------------------
+# Note: this is an experimental feature introduced in v 4.6.4a
+# It may be removed or changed in future releases
+
+scheme_formatters = {}
+for __p in pkg_resources.iter_entry_points('taurus.qt.qtgui.base.formatter'):
+    try:
+        scheme_formatters[__p.name] = '{}.{}'.format(__p.module_name,
+                                                     __p.attrs[0])
+    except Exception as e:
+        warning('Could not load formatter plugin "%s". Reason: %s',
+                __p.module_name, e)
+
+# ------------------------------------------------------------------------
 
 
 def defaultFormatter(dtype=None, basecomponent=None, **kwargs):
--- a/lib/taurus/qt/qtgui/base/taurusbase.py
+++ b/lib/taurus/qt/qtgui/base/taurusbase.py
@@ -69,22 +69,6 @@ DefaultNoneValue = "-----"
 
 
 
-# ------------------------------------------------------------------------
-# Note: this is an experimental feature introduced in v 4.6.4a
-# It may be removed or changed in future releases
-
-scheme_formatters = {}
-for __p in pkg_resources.iter_entry_points('taurus.qt.qtgui.base.formatter'):
-    try:
-        scheme_formatters[__p.name] = '{}.{}'.format(__p.module_name,
-                                                     __p.attrs[0])
-    except Exception as e:
-        warning('Could not load formatter plugin "%s". Reason: %s',
-                __p.module_name, e)
-
-# ------------------------------------------------------------------------
-
-
 def defaultFormatter(dtype=None, basecomponent=None, **kwargs):
     """
     Default formatter callable. Returns a format string based on dtype
@@ -99,6 +83,10 @@ def defaultFormatter(dtype=None, basecomponent=None, **kwargs):
     if issubclass(dtype, Enum):
         dtype = Enum
     return basecomponent.defaultFormatDict.get(dtype, "{0}")
+
+
+expFormatter = '{:2.3e}'
+floatFormatter = '{:.5f}'
 
 
 class TaurusBaseComponent(TaurusListener, BaseConfigurableClass):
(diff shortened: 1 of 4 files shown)


lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -66,6 +68,23 @@
...
+for __p in pkg_resources.iter_entry_points('taurus.qt.qtgui.base.formatter'):
+    try:
+        scheme_formatters[__p.name] = '{}.{}'.format(__p.module_name,
+                                                     __p.attrs[0])
+    except Exception as e:
+        warning('Could not load formatter plugin "%s". Reason: %s',
+                __p.module_name, e)
+
+# ------------------------------------------------------------------------
+
```
cpascual: In terms of implementation, I think that this is not the best way of collecting the registered formatters for the following reasons:
- I would call the entry point `taurus.qt.qtgui.base.formatters` (note the final "s"), or even `taurus.qt.formatters`
- The iteration and collection can be simplified (see proposed snippet below)
- The discovery is better delayed until the moment is needed (e.g. define getter for the formatters instead of loading them when importing the module).

In summary, I would do something more or less like this:
```python
def get_registered_formatters():
    formatters = {}
    for ep in pkg_resources.iter_entry_points('taurus.qt.qtgui.base.formatters'):
        try:
            formatters[ep.name] = ep.load()
        except Exception as e:
            taurus.warning('Cannot load "%s" formatter. Reason: %r', ep.name, e)
    return formatters
```

lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -1342,6 +1361,8 @@ def __init__(self, name='', parent=None, designMode=False, **kwargs):
         self.call__init__(TaurusBaseComponent, name,
                           parent=parent, designMode=designMode)
         self._setText = self._findSetTextMethod()
+        self._formaters = scheme_formatters
+        self._formaters['{:2.3e}'] = '{:2.3e}'
```
cpascual: spelling: use "formatter" instead of "formater" (check also other usages)

cpascual: Similar to what I was mentioning before: I would not keep a list of loaded formatters referenced by the TaurusBaseWidget. In other words, instead of keeping `self._formaters` I would just call the getter whenever is needed (e.g. in `showFormatterDlg()`)

cpascual: Instead of defining the exponential formatter here, I would simply register it using the entry point (just as the default formatter)

setup.py
```diff
@@ -118,10 +118,16 @@ def get_release_info():
...
+formatters = [
+    'default = taurus.qt.qtgui.base.taurusbase:defaultFormatter',
+    'tango = taurus.core.tango.util.formatter:tangoFormatter',
+]
+
 entry_points = {
     'console_scripts': console_scripts,
     'taurus.cli.subcommands': taurus_subcommands,
     'taurus.qt.qtgui.panel.TaurusModelSelector.items': model_selectors,
+    'taurus.qt.qtgui.base.formatter': formatters,
```
cpascual: As mentioned above, use "formatter**s**"

lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -1350,13 +1371,27 @@ def showFormatterDlg(self):
         (in string version) or None
         """
         current_format = self.getFormat()
+        ind = self._formaters.values().index(current_format)
+
+        formatter, ok = Qt.QInputDialog.getItem(self, "Set formatter",
```
cpascual: Since the dialog does not allow to customize the tooltip, use the "label text" to give the hint (maybe using an html link to the formatter docs?)

lib/taurus/qt/qtgui/base/taurusbase.py
cpascual: Note the getter could also be implemented as a static method of TaurusBaseComponent (to make it available to any widget

cmft:
Done

lib/taurus/qt/qtgui/base/taurusbase.py
cpascual: After discussing with @cmft , we agreed that keeping a list may still be needed in order to offer persistence in the choice of formatter hwen the user sets a custom formatter

lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -1342,21 +1366,52 @@ def __init__(self, name='', parent=None, designMode=False, **kwargs):
...
+            self._formatters = TaurusBaseWidget.get_registered_formatters()
+
+        # Find index of the current formatter
+        try:
+            ind = self._formatters.values().index(self.FORMAT)
+        except ValueError:
+            ind = 0
+
+        uri = "http://taurus-scada.org/devel/api/taurus/qt/qtgui/base/_TaurusBaseComponent.html"
+        help = "<a href=\"{}\"> Find more info in doc</a>".format(uri)
```
cpascual: Not crytical, but the link does not launch a browser in my system when clicked (but it allows to right-click and copy the URL which , when pasted on a browser works as expected)

lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -1342,21 +1366,52 @@ def __init__(self, name='', parent=None, designMode=False, **kwargs):
...
+
+        msg = "Choose/Enter a formatter: {}".format(help)
```
cpascual: I would replace `msg` and `help` by:

```python
msg = 'Choose/Enter a formatter (<a href="{}">help</a>)'.format(uri)
```

lib/taurus/qt/qtgui/base/taurusbase.py
```diff
@@ -1342,21 +1366,52 @@ def __init__(self, name='', parent=None, designMode=False, **kwargs):
...
         (in string version) or None
         """
-        current_format = self.getFormat()
+        if self._formatters is None:
+            # initialize cache
+            self._formatters = TaurusBaseWidget.get_registered_formatters()
+
+        # Find index of the current formatter
+        try:
+            ind = self._formatters.values().index(self.FORMAT)
```
cpascual: This fails with py3

cpascual: fixed in 5e32a69

lib/taurus/qt/qtgui/base/taurusbase.py
cpascual: fixed in 5e32a69

cpascual:
I did some heavy refactoring. Can some @taurus-org/integrators  test it?

Here is a snippet of a code that can be used to test:

```python
import taurus

def myformatter(*a, **kw):
    return "**{1.2e}**"

if __name__ == '__main__':
    import sys
    w = TaurusForm()
    w.setModifiableByUser(True)
    model = ['sys/tg_test/1/float_scalar', 'sys/tg_test/1/double_scalar']
    w.setModel(model)
    w.show()

    # w[0].setFormat(tangoFormatter)
    # w[0].setFormat('taurus.core.tango.util.tangoFormatter')
    # w[0].setFormat(myformatter)
    w[0].setFormat('///{:1.2f}///')

    # show setFormatter dialog
    w[0].onSetFormatter()
    sys.exit(app.exec_())
```

I just see one thing missing: if the selected formatter happens to be `None` (which is often the case for the class formatter), `onSetFormatter` does not set it. This requires either deprecating the `onSetFormatter`/`showFormatterDlg` API or breaking it. I do not expect anyone having reimplemented these methods in their own code, so I vote for breaking in this corner case

cpascual:
mmm... I [clearly broke the formatters](https://travis-ci.org/taurus-org/taurus/builds/653369204) with my refactoring... I'll investigate and fix

cpascual:
All clear now. Merging

PR #1057, opened 2020-01-30
(doc) Improve v3->v4 migration guide

cpascual:
The Migration guide suggests using TangoDevice.stateObj.read().rvalue as
a replacement for the deprecated TangoDevice.getState(), but the
behaviour is not identical since the suggested replacement uses cache.
Change docs to suggest a closer replacement.


reszelaz:
Thanks for the PR!

@cmft points that the `TangoDevice.getState` in taurus 3 was also returning the value from the cache by default. So, to be 100 % equivalent we should do the same with `TangoDevice.stateObj.read().rvalue`.

However there is another case:
`TaurusDevice.state()` which in taurus 3 used to work and was actully reading the "State" attribute via PyTango, now does not work. The migration guide should either propose:
`TaurusDevice.read_attribute("State").value` or eventually `TangoDevice.stateObj.read(cache=False).rvalue`. The first option will still use PyTango, the second option will use taurus but with the same result as when used with taurus 3.

cpascual:
Thanks @cmft and @reszelaz for pointing this out. I am closing this ticket as rejected and opening #1059  wich also fixes a related issue.

PR #1058, opened 2020-01-31
Fix missing precision initialization

cmft:
Some widgets report 'Invalid format' warning messages for
TangoAttributes URIS of Type.Integer.

The precision attribute is not initialized right for integer types.

Fix it.

You can reproduce it with : `taurus --log-level=Debug form sys/tg_test/1/long_scalar`

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -1078,7 +1078,7 @@ class TangoAttribute(TaurusAttribute):
         match = re.search("[^\.]*\.(?P<precision>[0-9]+)[eEfFgG%]", fmt)
         if match:
             self.precision = int(match.group(1))
-        elif re.match("%[0-9]*d", '%d'):
+        elif re.match("%[0-9]*d", fmt):
             self.precision = 0
         # self._units and self._display_format is to be used by
         # TangoAttrValue for performance reasons. Do not rely on it in other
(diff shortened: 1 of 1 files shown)


lib/taurus/core/tango/tangoattribute.py
```diff
@@ -1078,6 +1078,8 @@ def _decodeAttrInfoEx(self, pytango_attrinfoex=None):
         match = re.search("[^\.]*\.(?P<precision>[0-9]+)[eEfFgG%]", fmt)
         if match:
             self.precision = int(match.group(1))
+        elif re.match("%[0-9]*d", '%d'):
```
cpascual: I think that this line is wrong: it is static and always evaluates to `True `

Perhaps you intended to write the following?

```python
elif re.match("%[0-9]*d", fmt):`
```

cmft: You are right

cmft:
> Or to be more precise, I would like to understand why taurus --log-level=Debug form 'eval:Q(123)' does not trigger the same issue.

That is true. But in case that you do not set `precision` to `0` for int values (without mattering for which scheme). The `defaultFormatDict ` is wrong. Cause `int` and `float` attributes are represented by Quantities and we are saying that `float` has precision   and  `int` does not. 

IMO the bug is not triggered for `eval` for other reasons (as we see, it is not using Quantity to display the value) but is there.

See:
```
    defaultFormatDict = {float: "{:.{bc.modelObj.precision}f}",
                         Enum: "{0.name}",
                         Quantity: "{:~.{bc.modelObj.precision}f}"
                         }
```

cpascual:
The travis tests are failing for py2qt5, but this is already fixed in develop by #1075 , so I am merging.

PR #1059, opened 2020-02-03
Improve v3->v4 migration API and docs

cpascual:
- Fix an issue with bck-compat method `TangoDevice.getDisplayValue()`
- Improve v3->v4 migration guide regarding tango states

This PR replaces #1057

### DIFF
--- a/doc/source/devel/taurus3to4.rst
+++ b/doc/source/devel/taurus3to4.rst
@@ -154,7 +154,8 @@ This list is a best-effort to document changes, but it may not be 100% complete 
 | TangoDevice.getValueObj                       | TangoDevice.stateObj.read  *tango* or                         |
 |                                               | state  *agnostic*                                             |
 +-----------------------------------------------+---------------------------------------------------------------+
-| TangoDevice.getDisplayValue                   | TangoDevice.state().name                                      |
+| TangoDevice.getDisplayValue                   | TangoDevice.stateObj.read().rvalue.name  *tango*  or          |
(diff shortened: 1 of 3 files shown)


doc/source/devel/taurus3to4.rst
```diff
@@ -151,11 +151,11 @@ This list is a best-effort to document changes, but it may not be 100% complete
 +-----------------------------------------------+---------------------------------------------------------------+
 | TangoDevice.getSWState                        | TangoDevice.state                                             |
 +-----------------------------------------------+---------------------------------------------------------------+
-| TangoDevice.getValueObj                       | TangoDevice.stateObj.read  *tango* or                         |
-|                                               | state  *agnostic*                                             |
+| TangoDevice.getValueObj                       | TangoDevice.stateObj.read()  *tango* or                       |
```
cpascual: Hi @reszelaz, thanks for reviewing, but I do not think that this precise correction is right.
The replacement for the **method object** `TangoDevice.getValueObj` is the **method object** `TangoDevice.stateObj.read` (no parentheses)

This is the cryterium that we used when writing the table (although I am aware that we may not have been 100% consistent along the whole table)

reszelaz: Sorry, I was mislead that in the context of this change there were a lot of `read()` calls. Looking at the whole table I see the general cryterium. I just reverted this change.

cpascual:
Since this got the approval of @reszelaz , I am  merging it myself

PR #1063, opened 2020-02-10
Don't store classid config TaurusValue unless it is a string

### LINKED ISSUE #840: createConfig: LabelWidget not saved
Hi,

When I close taurusgui with many motors or just do Save Perspective, I get:

`MainThread     INFO     2019-01-17 15:14:32,764 PoolMotorTV.tango://b-v-balder-csdb-0.maxiv.lu.se:10000/b308a-o05/opt/slit-01-vb: createConfig: LabelWidget not saved because it is not Pickable (<class 'sardana.taurus.qt.qtgui.extra_pool.poolmotor.PoolMotorTVLabelWidget'>)`
for each motor -- a very long output.

Can you please look at it?

Taurus 4.4.0, Sardana 2.5.0
 
Thanks!

ByRellex:
TaurusValue refuses to store configs that have items without classID. This PR modifies the code to save those items leaving a debug message that tells the classID wasn't stored.
This PR fixes #840 which was actually a taurus issue, not a Sardana issue as indicated.
So also fixes https://github.com/sardana-org/sardana/issues/1164

### DIFF
--- a/lib/taurus/qt/qtgui/panel/taurusvalue.py
+++ b/lib/taurus/qt/qtgui/panel/taurusvalue.py
@@ -1165,13 +1165,14 @@ class TaurusValue(Qt.QWidget, TaurusBaseWidget):
                     or allowUnpickable):
                 #configdict[key] = classID
                 configdict[key] = {'classid': classID}
-                widget = getattr(self, key[0].lower() + key[1:])()
-                if isinstance(widget, BaseConfigurableClass):
-                    configdict[key]['delegate'] = widget.createConfig()
             else:
-                self.info('createConfig: %s not saved because it is not Pickable (%s)' % (
+                self.debug('createConfig: classid of %s not saved because it is not Pickable (%s)' % (
                     key, str(classID)))
-
+            widget = getattr(self, key[0].lower() + key[1:])()
+            if isinstance(widget, BaseConfigurableClass):
+                widget_configdict = configdict.get(key, {})
+                widget_configdict['delegate'] = widget.createConfig()
+                configdict[key] = widget_configdict
         return configdict
 
     def applyConfig(self, configdict, **kwargs):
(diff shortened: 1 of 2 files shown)


cpascual:
Hy @ByRellex , thanks for the proposed fix.
However, I lack context to understand the reason behind the changes. Could you please explain how these changes fix #840?

lib/taurus/qt/qtgui/panel/taurusvalue.py
```diff
@@ -1165,13 +1165,14 @@ def createConfig(self, allowUnpickable=False):
...
             else:
-                self.info('createConfig: %s not saved because it is not Pickable (%s)' % (
+                self.debug('createConfig: classid of %s not saved because it is not Pickable (%s)' % (
                     key, str(classID)))
-
+            widget = getattr(self, key[0].lower() + key[1:])()
+            if isinstance(widget, BaseConfigurableClass):
+                widget_configdict = configdict.get(key, {})
+                widget_configdict['delegate'] = widget.createConfig()
+                configdict[key] = widget_configdict
```
cpascual: I am not sure about this change: by moving the configdict creation outside the "if", potentially unpickable configs would be created **even when `allowUnpickable == False`**, which I do not think is desirable. 

If a given widget (e.g. `PoolMotorTV`) can handle unpickable configs, maybe the whole issue is solved by simply defining `allowUnpickable=True` for that widget...

cpascual: I correct myself: no unpickable configs would be created, but still there would be an issue when calling applyConfig with this config in some situations if classid is not stored.

cpascual:
After some discussion with @ByRellex and @reszelaz , we concluded that while this approach fixes some cases, it introduces a potential issue when changing the a TaurusValue subwidget class at runtime if classid property is not available.

A better approach is to support (de)serialization of classid in `TaurusValue.createConfig` and `TaurusValue.applyConfig` , so I am rejecting this PR in favour of #1066

PR #1072, opened 2020-02-18
Avoid NoneType AttributeError

### LINKED ISSUE #1071: PMTV in compact mode raises expception on Taurus GUI start
To reproduce:
1. Create taurus newgui
2. Add a TaurusForm panel to the GUI and set a motor as the model
3. Change PMTV to compact mode
4. Close the GUI
5. Open the GUI

Then you get:

```
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/base/taurusbase.py", line 394, in filterEvent
    self.handleEvent(*evt)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/input/tauruslineedit.py", line 151, in handleEvent
    self._updateValidator(evt_value)
  File "/home/zreszela/workspace/taurus/lib/taurus/qt/qtgui/input/tauruslineedit.py", line 98, in _updateValidator
    bottom, top = attr.range
AttributeError: 'NoneType' object has no attribute 'range'
```

cmft:
TaurusValueLineEdit._updateValidator method reads the range from
the model object and this can be None (not initialized).

Use the range of the given taurus value.

Fix #1071

### DIFF
--- a/lib/taurus/qt/qtgui/input/tauruslineedit.py
+++ b/lib/taurus/qt/qtgui/input/tauruslineedit.py
@@ -98,7 +98,7 @@ class TaurusValueLineEdit(Qt.QLineEdit, TaurusBaseWritableWidget):
             if attr is None:
                 bottom, top = -numpy.inf, numpy.inf
             else:
-                bottom, top= value.range
+                bottom, top = value.range
             if bottom != val.bottom:
                 val.setBottom(bottom)
             if top != val.top:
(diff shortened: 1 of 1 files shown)


cpascual:
Hi, thanks for the PR.
However, this is making the tests fail because it uses a deprecated API:

```
TangoAttrValue.range is deprecated since 4.0. Use TangoAttribute.range instead
```

PR #1087, opened 2020-03-02
Make sure the tooltip can be generated despite DevFail

### LINKED ISSUE #1086: terminate called after throwing an instance of 'Tango::DevFailed'
Hi,

I have created a very small Taurus application terminating this way.

```
terminate called after throwing an instance of 'Tango::DevFailed'
```
THis message is a guess print by Qt at the event loop termination. But i don't know.

I turn the python looging module to debug but there is nothing.

By chance do you have any idea how to debug this kind of things?

vallsv:
Here is a patch to allow to display tooltip for `DevFail ` attributes.

Screenshot from 2020-03-02 17-55-58

Without catching this exception, the `TaurusLabel` devices can't start. As result.

This patch is not really nice cause the patched method have to return str (i am afraid about returning something else), then the formatting have to be done inside `getDisplayDescrObj`

This could be improved by:
- Returning the list containing `value=exception`, or an object which mimicks `str`
- The formatting could then be done inside `TaurusBaseComponent.toolTipObjToStr`
- The tooltip rendering could be then improved a lot

Closes #1086

### DIFF
--- a/lib/taurus/core/tango/tangodevice.py
+++ b/lib/taurus/core/tango/tangodevice.py
@@ -185,8 +185,11 @@ class TangoDevice(TaurusDevice):
         ret = []
         for name, value in desc_obj:
             if name.lower() == u'device state' and self.stateObj is not None:
-                tg_state = self.stateObj.read(cache).rvalue.name
-                value = u"%s (%s)" % (value, tg_state)
+                try:
+                    tg_state = self.stateObj.read(cache).rvalue.name
+                    value = u"%s (%s)" % (value, tg_state)
+                except Exception as e:
+                    value = u"cannot read state"
             ret.append((name, value))
         return ret
 


lib/taurus/core/tango/tangodevice.py
```diff
@@ -185,9 +185,14 @@ def getDisplayDescrObj(self, cache=True):
         ret = []
         for name, value in desc_obj:
             if name.lower() == u'device state' and self.stateObj is not None:
-                tg_state = self.stateObj.read(cache).rvalue.name
-                value = u"%s (%s)" % (value, tg_state)
-            ret.append((name, value))
+                try:
```
cpascual: I would prefer to make this exception handling a bit simpler and more robust (even if sacrificing some info in the tooltip). I propose  to replace the try-except block by:

```python
    try:
        tg_state = self.stateObj.read(cache).rvalue.name
    except Exception as e:
        tg_state = "cannot read state"
    value = u"%s (%s)" % (value, tg_state)
```

lib/taurus/core/tango/tangodevice.py
```diff
@@ -185,9 +185,14 @@ def getDisplayDescrObj(self, cache=True):
...
+                    tg_state = self.stateObj.read(cache).rvalue.name
+                    value = u"%s (%s)" % (value, tg_state)
+                except DevFailed as e:
+                    error = str(e).replace(u"\n", u"<br/>")
+                    template = u"""%s <font style="white-space:pre">%s</font>"""
+                    value = template % (value, error)
+                ret.append((name, value))
```
cpascual: It looks like in your commit, you moved the `ret.append` inside the `if` block while originally it was in the outer `for` block (i.e. you added one indentation)

Was this on purpose? If so, why?

lib/taurus/core/tango/tangodevice.py
vallsv: I prefer not to hide any exception.

For example using `Expcetion` will hide typo in the code. And the content of `DevFail` can be important. But i guess it there is an exception there, it will also be catched somewhere else in the code.

But while it works, i am fine with your proposal.

lib/taurus/core/tango/tangodevice.py
vallsv: It was just a mistake. Thanks for the review.

vallsv:
Here is an update.

cpascual:
thanks a lot for the contribution @vallsv . Merged.

PR #1089, opened 2020-03-04
Do not require the use of cmd_line_parser=None in TaurusApplication

### LINKED ISSUE #1065: check the type of cmd_line_parser arg of TaurusApplication
TaurusApplication should raise a TypeError if cmd_line_parser is neither None or a parser returned by taurus.core.util.argparse.get_taurus_parser().

See context in #1064

cpascual:
TaurusApplication implicitly uses a `optparse.OptionParser` instance  if it does not receive an explicit `cmd_line_parser` keyword argument.  This is inconvenient because it forces the user to explicitly pass  `cmd_line_parser=None` when using other mechanisms such as `click` or `argparse` to parse CLI options.

Change this so that, by default `cmd_line_parser=None` is assumed if not explicitly passed, but allow the user to revert to the old behaviour by setting `IMPLICIT_OPTPARSE=True` in  `tauruscustomsettings`.

Also, fixes #1065

### DIFF
--- a/lib/taurus/qt/qtgui/application/taurusapplication.py
+++ b/lib/taurus/qt/qtgui/application/taurusapplication.py
@@ -211,19 +211,25 @@ class TaurusApplication(Qt.QApplication, Logger):
         if parser is None:
             p, opt, args = None, None, None
         else:
-            if parser.version is None and app_version:
-                # if the parser does not contain version information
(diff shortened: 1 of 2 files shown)


cpascual:
Hi @taurus-org/integrators , could you please have a look at this PR?

PR #1090, opened 2020-03-06
Lazy plugin import

### LINKED ISSUE #944: Use lazy imports when loading the taurus.qt.qtgui entry point
Currently, importing `taurus.qt.qtgui` triggers the load of any module registered into the `"taurus.qt.qtgui"` entry-point. See this snippet from `taurus.qt.qtgui.__init__.py`:

```python
for __p in pkg_resources.iter_entry_points('taurus.qt.qtgui'):
    try:
        __modname = '%s.%s' % (__name__, __p.name)
        __mod = __p.load()
        # Add it to the current module
        ...
```

This is less efficient and contrasts with how "native" submodules of `taurus.qt.qtgui`behave (e.g. importing  `taurus.qt.qtgui` does not import `taurus.qt.qtgui.display`, but it does import `taurus.qt.qtgui.tpg`)

One proposed solution (thanks to @tiagocoutinho !) for this would be to avoid loading the entry point and inserting instead a module object constructed with `types.ModuleType()` which receives the entry point and implements a `__getattr__` that loads the entry point only on the first attempt to access one of its members.

stanislaw55:
Hi Taurus Community

This should close #944 

How does it work:
During import, fake `LazyModule` object is created and put into `sys.modules` with proper name (of the actual plugin module) and reference to entry point. When attribute not present in this fake module is requested, `__getattr__` is invoked. It then loads actual module and replaces fake module with actual one in `sys.modules`. It is done that way to avoid problems that may or may not occur. This tricks just has to work once for each plugin module.
After that, attribute from plugin module is returned. That way actual module is present in `sys.modules` even if requested attribute name is wrong or attribute is not present.

Tested with `taurus_pyqtgraph` under Python 2.7

### DIFF
--- a/lib/taurus/qt/qtgui/__init__.py
+++ b/lib/taurus/qt/qtgui/__init__.py
@@ -63,12 +63,14 @@ class LazyModule(ModuleType):
         self.ep = entry_point
 
     def __getattr__(self, member):
+        import sys as __sys
         mod = self.ep.load()
         # Replace lazy module with actual module for package
-        setattr(sys.modules[self.__package__], self.__name__, mod)
+        setattr(__sys.modules[self.__package__], self.__name__, mod)
         # Replace lazy module with actual module in sys.modules
         modname = "%s.%s" % (self.__package__, self.__name__)
-        sys.modules[modname] = mod
+        __sys.modules[modname] = mod
+        del __sys
         return getattr(mod, member)
 
 # Discover the taurus.qt.qtgui plugins
--- a/lib/taurus/qt/qtgui/__init__.py
+++ b/lib/taurus/qt/qtgui/__init__.py
@@ -35,9 +35,10 @@ import os
 import sys
 import glob
 import pkg_resources
-from types import ModuleType
+
 from taurus import tauruscustomsettings as __S
 from taurus import info as __info
+from taurus.core.util.lazymodule import LazyModule
 
 
 __docformat__ = 'restructuredtext'
@@ -53,33 +54,6 @@ __icon.registerTheme(name=getattr(__S, 'QT_THEME_NAME', 'Tango'),
 # ------------------------------------------------------------------------
 # Note: this is an experimental feature introduced in v 4.3.0a
 # It may be removed or changed in future releases
-
-class LazyModule(ModuleType):
-
-    def __init__(self, name, package, entry_point):
-        super(LazyModule, self).__init__(name)
-        self.__package__ = package
-        self.ep = entry_point
-
-    def __getattr__(self, member):
-        import sys as __sys
-
-        from taurus import warning as __warning
-
-        try:
-            mod = self.ep.load()
-            # Replace lazy module with actual module for package
-            setattr(__sys.modules[self.__package__], self.__name__, mod)
-            # Replace lazy module with actual module in sys.modules
-            modname = "%s.%s" % (self.__package__, self.__name__)
-            __sys.modules[modname] = mod
-            return getattr(mod, member)
-        except Exception as e:
-            __warning('Could not load plugin "%s". Reason: %s', __p.module_name, e)
-            return None
-        finally:
-            del __sys, __warning
-
 
 # Discover the taurus.qt.qtgui plugins
 __mod = __modname = None
--- a/lib/taurus/qt/qtgui/__init__.py
+++ b/lib/taurus/qt/qtgui/__init__.py
@@ -38,7 +38,6 @@ import pkg_resources
 from types import ModuleType
 from taurus import tauruscustomsettings as __S
 from taurus import info as __info
-from taurus import warning as __warning
 
 
 __docformat__ = 'restructuredtext'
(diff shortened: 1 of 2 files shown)


lib/taurus/qt/qtgui/__init__.py
```diff
@@ -54,16 +55,34 @@
...
 
+class LazyModule(ModuleType):
+
+    def __init__(self, name, package, entry_point):
+        super(LazyModule, self).__init__(name)
+        self.__package__ = package
+        self.ep = entry_point
+
+    def __getattr__(self, member):
+        import sys as __sys
```
cpascual: why is this import (and later deletion) of `sys` needed? 
AFAIK, `sys` is already imported at global module level...

lib/taurus/qt/qtgui/__init__.py
```diff
@@ -54,16 +55,34 @@
...
+        mod = self.ep.load()
```
cpascual: I think that this should be protected, since it is the point at which problems typically occur (e.g. due to missing dependencies for the plugin).
In fact, I think that the try-except block of the original code makes more sense here than in its original place now that the loading is delayed

lib/taurus/qt/qtgui/__init__.py
stanislaw55: I think this is called later, when `sys` is already deleted. However, I'm not sure about that. Python complains about not founding module when there's not import

lib/taurus/qt/qtgui/__init__.py
stanislaw55: You're right, will change

lib/taurus/qt/qtgui/__init__.py
cpascual: Oh... I see... it is because of the last line in the __init__ that is cleaning the namespace

I do  not remember why exactly we bothered in cleaning the namespace in the case of the __init__. Maybe it had something to do with the autodocumentation...

lib/taurus/qt/qtgui/__init__.py
```diff
@@ -64,31 +63,36 @@ def __init__(self, name, package, entry_point):
...
+            setattr(__sys.modules[self.__package__], self.__name__, mod)
+            # Replace lazy module with actual module in sys.modules
+            modname = "%s.%s" % (self.__package__, self.__name__)
+            __sys.modules[modname] = mod
+            return getattr(mod, member)
+        except Exception as e:
+            __warning('Could not load plugin "%s". Reason: %s', __p.module_name, e)
+            return None
+        finally:
+            del __sys, __warning
```
cpascual: I think we can gain in clarity and resuability by moving the implementation of LazyModule into its own submodule (e.g. to `.../taurus/core/util/lazymodule.py`). This way it can be used for other entry points and we do not need to "pollute" the name space of this init.

lib/taurus/qt/qtgui/__init__.py
```diff
@@ -64,31 +63,36 @@ def __init__(self, name, package, entry_point):
...
-        __info('Plugin "%s" loaded as "%s"', __p.module_name, __modname)
-    except Exception as e:
-        __warning('Could not load plugin "%s". Reason: %s', __p.module_name, e)
+    __modname = '%s.%s' % (__name__, __p.name)
+    __lazy_mod = LazyModule(__p.name, __name__, __p)
+    # Add it to the current module
+    setattr(sys.modules[__name__], __p.name, __lazy_mod)
+    # Add it to sys.modules
+    sys.modules[__modname] = __lazy_mod
+    __info('Plugin "%s" loaded as "%s"', __p.module_name, __modname)
```
cpascual: Maybe this could be implemented in a (static) method of `LazyModule` which would be called from here

cpascual: IMHO, the info message should say "lazy-loaded" instead of "loaded"

lib/taurus/qt/qtgui/__init__.py
cpascual: ... and in that case there is no need of worrying about cleaning the namespace (and more natural import names could be used , e.g. `sys` instead of `__sys`, etc.)

lib/taurus/qt/qtgui/__init__.py
```diff
@@ -54,22 +54,45 @@
 # Note: this is an experimental feature introduced in v 4.3.0a
 # It may be removed or changed in future releases
 
+class LazyModule(ModuleType):
+
+    def __init__(self, name, package, entry_point):
+        super(LazyModule, self).__init__(name)
+        self.__package__ = package
```
cpascual: Do we really need to overwrite the  `__package__` special attribute of `LazyModule`? 

Unless for some reason we want to overwrite, I think it is safer not to, and just store our "target" package name in `self._target_pkg_name` or something similar.

stanislaw55: I added it based on information [here](https://docs.python.org/3.6/library/types.html#types.ModuleType). However, this may not be needed as I did not tested it without it

lib/taurus/qt/qtgui/__init__.py
stanislaw55: Personally I would leave it as it is now. However, if you insist, I will change it.

cpascual: Ok. It is not critical IMHO. I just mentioned it because I think that the 3 lines are not obvious and that if we are to eventually reuse this, they could be implemented in just one place. but do not worry. I mark it as resolved.

lib/taurus/qt/qtgui/__init__.py
cpascual: Oh. I was not aware of this. Thanks for pointing to the docs. I agree with your implementation.

cpascual:
Ok. With the latest changes it looks good to me. Now I need to actually test it.

By the way: would it be possible for me to get write permission to modify this branch? 
I would like to add some unit tests and I would need to do it via PR to this branch, which is a bit cumbersome. 
Usually there is an option when creating a PR to allow the integrators to modify the PR branch (checked by default)  which I find quite useful to simplify the review process.

stanislaw55:
Hi @cpascual 
About permission: I do not know why but there's no checkbox or any option for me to check for allowing modifications into PR source branch by upstream repo maintainers. I search whole organisation and repository settings and I could not find anything to make it work. I'm truly sorry for the inconvenience but I do not know where to look for enabling it.

cpascual:
I used this for testing, with only one plugin (taurus_pyqtgraph)

```
from time import time

t1 = time()
import taurus.qt.qtgui
t2 = time()
from taurus.qt.qtgui import tpg
t3 = time()
a = tpg.TaurusPlot
t4 = time()

print("total=", t4-t1)
print("partial=", t2-t1, t3-t2, t4-t3)
```

Without lazy import:

```
MainThread     INFO     2020-05-07 17:07:04,028 TaurusRootLogger: Plugin "taurus_pyqtgraph" loaded as "taurus.qt.qtgui.tpg"
total= 0.9116969108581543
partial= 0.9116919040679932 4.291534423828125e-06 7.152557373046875e-07
```

With lazy import, the total amount of time is similar, but we see that the loading of the plugin got delayed (as intended):
```
MainThread     INFO     2020-05-07 17:07:58,242 TaurusRootLogger: Plugin "taurus_pyqtgraph" lazy-loaded as "taurus.qt.qtgui.tpg"
total= 0.9149420261383057
partial= 0.7525215148925781 5.9604644775390625e-06 0.16241455078125
```

PR #1091, opened 2020-03-18
WIP: Delegate unsubscribe from events to Tango C++ (PyTango#292)

reszelaz:
Tango C++ unsubscribe from any subscribed event in the DeviceProxy destructor. As workaround for PyTango#292 do not call unsubscribes in Taurus. Just mark the event type as unsubscribed so eventual event callbacks are ignored in push_event.

This PR is one of the options to workaround https://github.com/tango-controls/pytango/issues/292. It still does not work properly. I found the Sardana testsuite hung in the following situation" [no_unsub_bt.txt](https://github.com/taurus-org/taurus/files/4350042/no_unsub_bt.txt) and [no_unsub_py-bt.txt](https://github.com/taurus-org/taurus/files/4350043/no_unsub_py-bt.txt)

. I plan to push another option with delegating the unsubscribe to a worker thread.

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -287,6 +287,7 @@ class TangoAttribute(TaurusAttribute):
         # current event subscription state
         self.__subscription_state = SubscriptionState.Unsubscribed
         self.__subscription_event = threading.Event()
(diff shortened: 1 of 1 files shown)


reszelaz:
The root issue was solved in tango-controls/cppTango#699.

PR #1093, opened 2020-03-20
WIP: Usubscribe from events in worker thread (PyTango#292)

reszelaz:
Similarly to https://github.com/taurus-org/taurus/pull/1091, this PR tries to workaround PyTango#292. 
This one implements the 3.b option proposed by @schooft in https://github.com/tango-controls/pytango/issues/292#issuecomment-585652804. Thanks!

I tried this PR with https://github.com/sardana-org/sardana/pull/1312 ( which reverts the Sardana workaround and use _zombie_ concept in tests) and it worked for 7 testsuite executions already. Note that the usage of zombie API in tests most probably won't be need if we implement https://github.com/taurus-org/taurus/issues/1092.

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -82,7 +82,7 @@ def _unsubscribe_event(dev_proxy, event_id):
             # it unsubscribed from events itself
             pass
         else:
-            debug("Error trying to unsubscribe configuration events")
+            debug("Error trying to unsubscribe events")
             trace(str(df))
 
 
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -50,6 +50,7 @@ from taurus.core.taurusoperation import WriteAttrOperation
 from taurus.core.util.event import EventListener, _BoundMethodWeakrefWithCall
 from taurus.core.util.log import (debug, taurus4_deprecation,
                                   deprecation_decorator)
+from taurus.core.util.threadpool import ThreadPool
 
 from taurus.core.tango.enums import (EVENT_TO_POLLING_EXCEPTIONS,
                                      FROM_TANGO_TO_NUMPY_TYPE,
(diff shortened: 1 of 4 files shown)


cpascual:
This is making 2 tests to fail:test_cleanup_after_polling`

reszelaz:
> This is making 2 tests to fail:
> 
> * test_cleanup_after_polling
> * test_cleanup_state_after_polling

These most probably fail cause I commented the disabling of polling in [`TangoAttribute._finalUnsubscribeChangeEvents`](https://github.com/reszelaz/taurus/blob/cd22cc90bfab8a8c34deace573d4d46fca2fd7fc/lib/taurus/core/tango/tangoattribute.py#L788) which was not clear to me that it had to be done exactly there. Why not in `TangoAttribute.cleanUp`?

reszelaz:
Anyway I have tested it with a typical macro which discovered PyTango#292:

```python
import taurus
from sardana.macroserver.macro import macro


@macro()
def test_dev_gc(self):
    pool = self.door.get_pools()[0]
    moveables = []
    moveables.extend(pool.getElementNamesOfType("motor"))
    moveables.extend(pool.getElementNamesOfType("pseudomotor"))
    channels = []
    channels.extend(pool.getElementNamesOfType("ctexpchannel"))
    channels.extend(pool.getElementNamesOfType("zerodexpchannel"))
    channels.extend(pool.getElementNamesOfType("onedexpchannel"))
    for m in moveables:
        self.print("{}: {}".format(m, taurus.Device(m).getPosition()))
    for ch in channels:
        self.print("{}: {}".format(ch, taurus.Device(ch).getValue()))
```

And this fails. So 7 consecutive executions of the sardanatestsuite is not an indicator that this workaround works. The above macro is a better way to check it.

So, while executing the macro, the MacroServer hanged with:  [work_th_bt.txt](https://github.com/taurus-org/taurus/files/4362911/work_th_bt.txt) and [work_th_py-bt.txt](https://github.com/taurus-org/taurus/files/4362912/work_th_py-bt.txt).

reszelaz:
The root issue was solved in tango-controls/cppTango#699.

PR #1094, opened 2020-03-20
Add conda recipe

tiagocoutinho:
This adds a conda recipe to be able to build taurus conda packages.

The recipe fetches a specific version taurus from pypi to build the conda package.

Maybe for CI it would make sense to add another meta.yaml that builds against the local filesystem instead. Let me know if you find it useful. I already have one skeleton that I prepared for this case

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,0 +1,71 @@
+{% set name = "taurus" %}
+{% set version = "4.6.1" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
+
+build:
+  number: 1
+  noarch: python
+  entry_points:
+    - taurus = taurus.cli:main
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - pyqt
+    - lxml
+    - guidata
+    - guiqwt
+    - ipython
+    - pillow
+    - ply
+    - pythonqwt
+    - click
+    - future
+    - numpy
+    - scipy
+    - pymca
+    - pint
+    - python
+    - setuptools
+  run:
+    - pyqt
+    - lxml
+    - guidata
+    - guiqwt
+    - ipython
+    - pillow
+    - ply
+    - pythonqwt
+    - click
+    - future
+    - numpy
+    - scipy
+    - pymca
+    - pint
+    - python
+    - setuptools
+
+test:
+  imports:
+    - taurus
+    - taurus.core
+  commands:
+    - taurus --help
+
+about:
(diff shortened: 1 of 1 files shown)


cpascual:
> Maybe for CI it would make sense to add another meta.yaml that builds against the local filesystem instead. Let me know if you find it useful. I already have one skeleton that I prepared for this case

Yes, it would definitely be interesting.

cpascual:
and deployment to a conda forge channel would be awesome

But this can be done also on our internal gitlab CI that we use for debian packaging

tiagocoutinho:
> and deployment to a conda forge channel would be awesome

putting it into conda-forge channel requires following an acceptance process with a probation period.
I have never done this so I don't know the details. My feeling is that it is too much trouble and you gain almost nothing.
I propose we place it in the "alba-controls" channel for now.
The only difference is that in the instructions we say:
`conda install taurus -c alba-controls`
instead of:
`conda install taurus -c conda-forge`
  
> But this can be done also on our internal gitlab CI that we use for debian packaging

Agree, I think the gitlab CI is the right place for it.
We probably need to:
1. generate a token on the channel we want to publish (ex: [alba-controls](https://anaconda.org/ALBA-controls/settings/access)).
2. register the token in gitlab as a variable
3. create the recipe which basically does: `anaconda upload -u <channel> <build file>`

cpascual:
> I propose we place it in the "alba-controls" channel

Of course, you are right . I was actually thinking on a "conda channel" , not "conda forge channel" but I wrote it wrongly. I agree alba-controls may be a good place for it

Regarding how to implement CI, if we use our internal gitlab-CI: any suggestion on which image to use for building the conda package? 
maybe one from https://hub.docker.com/u/continuumio ?

cpascual:
Also: if we implement CI/CD conda packaging in the packaging gitlab-CI, do you think it is worth merging this PR anyway?

cpascual:
Hi @tiagocoutinho , would it be possible for you to upload the meta.yaml that builds against the local filesystem ?

Also (or maybe, alternatively), it may make sense to automatically update the hardcoded version in the meta.yml file via bumpversion.

tiagocoutinho:
Sure, no problem.
@rhomspuron found/improved a nice template which is I should use to upload here instead of the hack I did. 
It is quite elegant since most of it is agnostic to the package it is packaging (if the package fits the family of packages it was designed for: pure python, with semantic versioning, ...) which I think taurus fits well.

cpascual:
Hi, I just pushed deee234 as part of  the preparation of the Jul20 release (#1127). It adds a `conda/meta.yaml` based on the recipe from @rhomspuron.

Therefore I am closing this PR since it will be obsoleted by the next release.

cpascual:
Note: if I manage to do it, I'll try to include it also an auto-deployment by travis

cpascual:
Yup!! I managed to automate the conda build and deployment (using github workflows), and also made it general for other projects to use the same configuration

Issue #1042, opened 2019-12-11
CI showing random segfaults


cpascual:
I am concerned about some segfaults that we are [lately seeing in the tests](https://travis-ci.org/taurus-org/taurus/jobs/623565096#L472) . The traces [seem to blame (py?)tango's  event subscription](https://travis-ci.org/taurus-org/taurus/jobs/623565096#L523) (tox reports using pytango 9.3.1 from conda).

The segfault is not happening all the time, but it is worrying enough.

Hint: If we know of some some specific version of (py)tango (or other related lib) that is not affected, maybe we could try to pin the safe version in the `tox.ini`

cpascual:
Now that  tango-controls/cppTango#699 is merged, we expect that some of the instability problems in our CI can be solved if we use cppTango >= 9.3.4rc5 in our CI.

Currently [we pinned 9.3.3](https://github.com/taurus-org/taurus/pull/1050) in travis. 

Note: 
The tango docker images from https://github.com/tango-controls/tango-cs-docker seem to be no longer updated. Therefore we could consider moving to the docker images produced by SKA: https://github.com/ska-telescope/ska-docker

cpascual:
Note to myself: check the possibility that the paralellization of the jobs is causing errors (maybe because different parallel jobs use the same tango network name?)


This idea comes from the following note [here](https://github.com/ska-telescope/ska-docker/blob/master/docker-compose/Makefile):

> Never use the network=host mode when running CI jobs, and add extra
> distinguishing identifiers to the network name and container names to
> prevent collisions with jobs from the same project running at the same
> time.

cpascual:
#1109 seems to fix this

PR #1095, opened 2020-03-21
WIP: Usubscribe from events on the next subscription (PyTango#292)

reszelaz:
Similarly to #1091 and #1093, this PR tries to workaround https://github.com/tango-controls/pytango/issues/292.

I tried this PR with [sardana-org/sardana#1312](https://github.com/sardana-org/sardana/pull/1312) ( which reverts the Sardana workaround and use zombie concept in tests) and it works corrrectly for the sardanatestsuite and for the stress macro posted in https://github.com/taurus-org/taurus/pull/1093#issuecomment-602042643. So, IMO it is a satisfactory workaround.

Note that it has one side effect - it keeps garbage collected attributes subscribed until some new attribute is subscribed. To mitigate this effect attributes are marked as zombies - their push event return immediately. Also this should happen only for the configuration events - the change events should be unsubscribed with the last `TaurusModel.removeListener` call. Also if one wants to increase the frequency of unsubscribes the `tangoattribute._empty_unsub_queue()` could be called more frequently.

### DIFF
--- a/lib/taurus/core/tango/tangoattribute.py
+++ b/lib/taurus/core/tango/tangoattribute.py
@@ -36,6 +36,7 @@ import weakref
 import PyTango
 import numpy
 from functools import partial
+from queue import Queue
 
 from taurus import Manager
 from taurus.core.units import Quantity, UR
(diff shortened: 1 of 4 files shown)


reszelaz:
One could also try to play manually with GC. Something like:
* call `gc.collect()` before we subscribe to events. This way we will ensure that there are no pending unsubscribes to be done.
* disable GC before subscribe and enable it after

(I have not tried neither of them)

reszelaz:
The root issue was solved in https://github.com/tango-controls/cppTango/pull/699.

cpascual:
Thanks @reszelaz for taking care of this, and sorry for not having been as responsive as I should (I have been busy in other fronts).

One thing that remains to do is using the cppTango > 9.3.4rc5 in our CI. This could be one way to try to fix #1042

PR #1098, opened 2020-04-03
Fix #1097: Tango extension image devices resubscribe at every frame

### LINKED ISSUE #1097: Tango extension image devices resubscribe at every frame
The tango extension `taurus.core.tango.img` devices at each frame received subscribe and unsubscribe to configuration events.
In the frame handler code there is a temporary access to the image attribute which triggers this call.
As frames can arrive at a frequent pace this slows down considerable and, more importantly, triggers the tango-controls/pytango#292.

The fix should be simple: just keep a reference to the attribute in the device. I can make a PR

tiagocoutinho:
Keep a reference to the image attribute.
This avoids subscribe/unsubscribe event calls at each frame


cpascual:
only py27qt4 test failing (unrelated causes), so I am merging

PR #1102, opened 2020-04-14
Catch RuntimeError in tauruscombobox postDetach

### LINKED ISSUE #1100: Setting model for tauruscombobox raises RuntimeError with PySide2
It looks like PyQt5 raises `TypeError`  on disconnect if signal wasn't connected before, which is caught, but PySide2 raises `RuntimeError` and stops the application in this case.

> venv/lib64/python3.6/site-packages/taurus/qt/qtgui/input/tauruscombobox.py", line 78, in postDetach
>     self.currentIndexChanged.disconnect(self.writeIndexValue)
> RuntimeError: Failed to disconnect signal currentIndexChanged(int).

MikeFalowski:
In case of disconnecting signal which was not connected before, PyQt5 raises `TypeError`, but PySide2 raises `RuntimeError`.

This is the simplest solution, I guess. I wasn't sure if I should catch only `TypeError` when PyQt5 is used and only `RuntimeError` when PySide2 is used, so I omitted this.

Fix  #1100


cpascual:
Merged. Thanks for the contribution @MikeFalowski .

PR #1105, opened 2020-05-07
 Add expiration_period arg to TangoAttribute.read()

reszelaz:
There is no way different then modifying the polling period to make the attribute readout return the value of the last event. Add expiration period concept to the attribute single readout.

Note that it might also have sense to add a global experation period for the attribute so one would not need to specify it at every readout but I think in some special occasion it still would make sense to pass it as argument to the read method. So, for the moment I just add this one since this is what we need in Sardana.


cpascual:
LGTM, but, have you considered using the `cache` arg for the expiration date? the algorithm would be:

```python
def read(self, cache=True):
   ...
   if cache:
       ...
       if cache is True:  # cache *is* a bool True
           expiration_period = self.getPollingPeriod()
       else:  # cache is a non-zero numeric value
           expiration_period = cache  
    if dt < expiration_period:
        ...
```   

The advantage that I see to this proposal is that you keep the number of args the same as in the other schemes (and to me it just feels ok to pass a possitive value as cache... actually it even allows to pass it a negative number to force cache no matter what)

reszelaz:
Ok, I like this idea. I have made the necessary changes. Also I changed the time unit to milliseconds to be coherent with the `changePollingPeriod()` (even if I would prefer seconds:))

PR #1112, opened 2020-05-18
Allow any case in CLI --log option

cpascual:
The l--taurus-log-level option of old CLI scripts was replaced by the 
--log option to the taurus script (#856). The old one was case-insensitive, while the new one is not.. Make the new one insensitive. For example, allow :
```
# only this is valid before this PR
taurus --log-level=Debug form

# this should be also valid now
taurus --log-level=debug form
```

### DIFF
--- a/lib/taurus/cli/common.py
+++ b/lib/taurus/cli/common.py
@@ -85,11 +85,16 @@ script::
 import click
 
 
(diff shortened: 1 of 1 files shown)


cpascual:
> This kwarg is not accepted by click 6.6 (Debian strech). Do we want to raise the click requirement?

Ups... thats a very good point!
No, I would not raise the requirement above the stretch defaults just for this. I'll propose a conditional implementation

PR #1114, opened 2020-05-27
Fix test suite

cpascual:
The test suite is still too fuzzy, and the [py35-qt5 target had to be quarantined](https://github.com/taurus-org/taurus/pull/1108#issuecomment-631728017).

This PR does a large refactoring of the tests (mostly the `taurus/qt` ones) to improve the situation. With it, the testsuite seems finally stable for all the targets. As a summary, here is what changed:

- refactor all the qt tests to use pytest-qt (this seemed to fix many of the "random" crashes during test execution). And deprecate old unittest-based qt test utilities and TimeOut DS.
- force to use PyTango >=9.3.1 for all targets (I had to provide my own conda package of PyTango v9.3.1 which was not officially available, see ab6888b )
- Add some pytest-based utilities for new tests
- Fix a few bugs in the code that affect the tests (8efbe87 , 14d94b8, e48a2d9, dcb6831,  )
- Remove taurus usages of deprecated APIs of dependencies (1d44c7a, 46ddb2c, b5173ad8, 964b18fa, 6cb42ff, 4aaf110a, 5bc24ad) and configure the tests to treat future occurrences as errors (66e4d32)
- change the tango docker images to use a mirror from those of ska-docker to avoid [availability issues of ska's registry](https://github.com/ska-telescope/ska-docker/issues/16)

Also, this PR removes the allowed-to-fail tests (python2 ones and flake8) from travis matrix (to avoid wasting time). They can be run locally and/or re-enabled at any moment

PR #1116, opened 2020-06-13
Config loaders

stanislaw47:
Hello Taurus community
This PR provides what I wanted in Taurus for long time: ability for load configuration for TaurusGUI from JSON.

I created abstraction for config source, be it a Python module or XML/JSON file. In that way, adding support for any other config file format, like TOML or YAML should be trivial. TaurusGUI now loads configuration by accessing fields of `*ConfigLoader` class. Whole interface of `*ConfigLoader` class is defined in `AbstractConfigLoader` by using Python `abc` module.

All discovery logic happens in `taurus.qt.qtgui.taurusgui.config_loader.getLoader` function. It matches loaders by file extensions. In case of directory or non-existing path, it tries to use `PyConfigLoader`. In case of using entry-points, logic has to be moved elsewhere but for the moment I do not know how to achieve it, especially when taking into account possibility to pass directory and Python module path. 

Regarding `abc.ABC` class for Python 2 and 3 - I know that Taurus uses `future` package for compatibility. However, I did not found anything there about providing `abc.ABC` so I created my own compatibility from what I found on StackOverflow.

Things to do:
 - [ ] test for loading configuration from different sources
 - [ ] backward compatibility tests
 - [x] entry-point for registering custom config loaders


cpascual:
Hi @stanislaw47 Thanks a lot for this contribution!, here are a few preliminary comments:

- I really like the idea, and I am happy because it comes to clean quite a messy part of Taurus (the config loading)
- Unless I missed something (I gave a quick glance only) , I thing this PR still misses the part in which the `getLoader` is called, and then the returner loaded used, right?
- I agree in that `getLoader` should be implemented with entry-points. I think it is relatively easy (I've been doing some work in preparation of [TEP13](https://github.com/taurus-org/taurus/blob/develop/doc/source/tep/TEP13.md) lately and I can help with it. I suggest to use the following entry point name: `taurus.gui.loaders` and iterate over it in `getLoader`. Also, the `*ConfigLoader` itself can declare its supported extension(s) (and we treat the pyconf case as an special fallback case)
- As for the abstract loader: I think this is a good opportunity to clean old/deprecated stuff. For example, the monitor stuff can be removed. Also, the sardana-related configs should not be implemented/required in taurus. A possible solution is to allow multiple config loaders for the same extension (e.g. if you provide a json file, all the loaders that declare themselves supporting json would be activated and used, and in this way taurus would declare a json loader that handles onlyy taurus configs while sardana would declare a json loader that handles the sardana-related configs,  and both would be used). Note that in order to completely isolate taurus and sardana configuration handling, the `TaurusGui.loadConfiguration()` will also need to be refactored to use plugins, but this can be done in two steps.
- Regarding python2 compatibility: I do not expect to maintain py2 support in taurus after the Jul20 release, so I personally would not struggle with making the new feature py2-compatible. If it is just a matter of the small compatibility hack with abc, then it is fine, but if it requires more work, I would make this a py3-only feature and fall back to the old mechanism for py2 if we integrate this PR before Jul20. The same for Qt4.
- Regarding tests: I just did a large refactoring of the tests in #1114 . As soon as it is merged, you should update this branch with it. Then I would use pytest to do unit tests for the loaders and for the backwards-compatibility
- Do you have any plans for a "*ConfigWriter?". At the moment, both the "new gui wizard" and the TaurusGui themselves are able to create xml config files.

stanislaw47:
Hi @cpascual 
> I really like the idea, and I am happy because it comes to clean quite a messy part of Taurus (the config loading)

I'm happy to hear it

> Unless I missed something (I gave a quick glance only) , I thing this PR still misses the part in which the getLoader is called, and then the returner loaded used, right?

Actually, it's there, in [here](https://github.com/stanislaw47/taurus/blob/8b857505c3d91586300cd471274732943356ccca/lib/taurus/qt/qtgui/taurusgui/taurusgui.py#L990). Resulting `conf` object is instance of `*ConfigLoader` nad is passed to specific loading methods.

> I agree in that getLoader should be implemented with entry-points. I think it is relatively easy (I've been doing some work in preparation of TEP13 lately and I can help with it. I suggest to use the following entry point name: taurus.gui.loaders and iterate over it in getLoader. Also, the *ConfigLoader itself can declare its supported extension(s) (and we treat the pyconf case as an special fallback case)

I got trouble getting my head around pyconf since it support noto only extension but also directories and non-exisitng paths. But if it'll be treated as special fallback case, then I would be trivial indeed. I'll try to implement it soon. 

> As for the abstract loader: I think this is a good opportunity to clean old/deprecated stuff. For example, the monitor stuff can be removed. Also, the sardana-related configs should not be implemented/required in taurus. A possible solution is to allow multiple config loaders for the same extension (e.g. if you provide a json file, all the loaders that declare themselves supporting json would be activated and used, and in this way taurus would declare a json loader that handles onlyy taurus configs while sardana would declare a json loader that handles the sardana-related configs, and both would be used). Note that in order to completely isolate taurus and sardana configuration handling, the TaurusGui.loadConfiguration() will also need to be refactored to use plugins, but this can be done in two steps.

I am happy to remove deprecated stuff like `MONITOR` or `CONSOLE`, I kept them for backward compatibility. About multiple loaders for same file extensions - I actually do not know how that would work with entrypoints. Would one plugin override the other? Or would it allow them to work at the same time? Could you describe it a little more?

> Regarding python2 compatibility: I do not expect to maintain py2 support in taurus after the Jul20 release, so I personally would not struggle with making the new feature py2-compatible. If it is just a matter of the small compatibility hack with abc, then it is fine, but if it requires more work, I would make this a py3-only feature and fall back to the old mechanism for py2 if we integrate this PR before Jul20. The same for Qt4.

For now, only compatibility is indeed `ABC` hack so I'll just keep it compatible for both Python 2 and 3. I believe this PR does not require special care for PyQt version, at least for now. 

> Regarding tests: I just did a large refactoring of the tests in #1114 . As soon as it is merged, you should update this branch with it. Then I would use pytest to do unit tests for the loaders and for the backwards-compatibility

Sure think, will do. I've written some tests using pytest so that should not be a problem. 

> Do you have any plans for a "*ConfigWriter?". At the moment, both the "new gui wizard" and the TaurusGui themselves are able to create xml config files.

Dumping configuration in given format would be cool feature. However, I think that might be non-trivial for py files. And could make this PR much bigger. Regarding plugins - should they support both loading and dumping configuration? Or just loading should be mandatory and dumping a feature? I personally would go for both loading and dumping as mandatory.

cpascual:
> About multiple loaders for same file extensions - I actually do not know how that would work with entrypoints. Would one plugin override the other? Or would it allow them to work at the same time? Could you describe it a little more?

What I had in mind is: the taurus abstract loader only forces the strictly-taurus configs. Then sardana would define a similar abstract loader which provides the same interface but forces a complementary set of configs.

Now, anyone (taurus or a third-party) can register any number of ConfigLoaders. Each ConfigLoader should provide a `.supports()` method which accepts the config source id (name of the config file/dir/module) and returns True/False depending if it supports it (in this way we do not need to rely just on extensions, and pyconf gets supported just as any other loader). 

When `getLoader` is called, instead of returning a single `*ConfigLoader`, it returns the list of those for which `support(id)` returned `True` . The list can be sorted for precedence in case of conflict (see note 1)

Then when loading the configs, all the ConfigLoaders would be called in order and allowed to update a config dict.

By the way: this method would eventually allow mixing config sources (e.g. it would be easy to naturally emulate the current situation in which configs are simultaneously loaded from a module and from an xml file)

Note 1: we can use the same mechanism as for the TaurusForm Item factory plugins to set the sort order

Note 2: (just a wild idea) maybe the ConfigLoader objects could present a dictionary-like API: this would make them quite natural to handle for operations such as "updating". It also would facilitate mocking for tests.

stanislaw47:
I've initially implemented plugable loader by using file extensions and pyconf as a special fallback (not pushed yet). However, using `supports` method would make it easier. One disadvantage I can see is that it makes each loader actually loaded from entrypoint. That would (probably) make startup slower, however I do not know if it would be significant for users.

Using `supports` would also make loading pyconf a bit more readable so obvious benefit here. 

> Then when loading the configs, all the ConfigLoaders would be called in order and allowed to update a config dict.

That would require a bit more refactoring. Now, configuration data is stored in properties of concrete `ConfigLoader`. There's no notion of 'config dict'. Using multiple config loaders as it is now would override one another completely. And if some option is not present in second one, default value provided by `TaurusGui.loadConfiguration` method would override value from first one. 
So basically:
There would be needed for one more object, something like `Configuration` class. It could expose dict-like API. Each loader would just extract data from its source and pass it to `Configuration` object. 
I think it has to have good architecture since there's high chance for repeating logic and managing data from wrong end. 

> By the way: this method would eventually allow mixing config sources (e.g. it would be easy to naturally emulate the current situation in which configs are simultaneously loaded from a module and from an xml file)

Hmm, not sure how it would *exactly* work with both Python module and XML file but I feel that it would be good way for supporting old configurations in backwards-compatible way.

lib/taurus/qt/qtgui/taurusgui/config_loader/abstract.py
```diff
@@ -0,0 +1,180 @@
...
+        return None
+
+    @abc.abstractmethod
+    def load(self):
+        """
+        This method is meant to load actual data from file on disk
+        """
+
+    @abc.abstractproperty
+    def conf_dir(self):
```
cpascual: Is this really necessary? It constrains the loaders to be file-based. I think that with a bit of care it should be easy to leave the possibility open to e.g. database-based config loaders

lib/taurus/qt/qtgui/taurusgui/config_loader/abstract.py
```diff
@@ -0,0 +1,180 @@
...
+        return []
+
+    @abc.abstractproperty
+    def external_apps(self):
+        """List of external applications"""
+        return []
+
+    # deprecated
+    @abc.abstractproperty
+    def monitor(self):
```
cpascual: As already discussed, this property (and maybe other deprecated ones) could be left out of the mandatory API. In fact, since we allow multiple loaders, they could be moved to a "bck-compat" type of loader

setup.py
```diff
@@ -125,11 +125,18 @@ def get_release_info():
     '{:.5f} = taurus.qt.qtgui.base:floatFormatter',
 ]
 
+config_loaders = [
```
cpascual, line 139: For the entry point name, I think it is better to use something that identifies the registrar (e.g., in this case `"taurus.py"`, `"taurus.json"`, `"taurus.xml"`, ...) because it helps for providing a mechanism to select/deselect the plugin (using `taurus.core.util.plugin.selectEntryPoints`).

cpascual:
>  using supports method would make it easier. One disadvantage I can see is that it makes each loader actually loaded from entrypoint. That would (probably) make startup slower, however I do not know if it would be significant for users.

Yes. I also thought about that, but the only alternative is using the registered plugin name for selection, and this would make it more difficult/ugly for allowing multiple loaders for the same type of source.

On the other hand, if we use the `taurus.core.util.selectEntryPoints` API, we can mitigate this by pre-filtering the entry point candidates if needed.

cpascual:
> There would be needed for one more object, something like Configuration class. It could expose dict-like API. Each loader would just extract data from its source and pass it to Configuration object.

Yes, this is more or less what I had in mind when I talked about "updating a config dict". I was imagining this Configuration object to be just a dict and then having the extra required logic in the loadConfiguration part, ... but probably using a Configuration class to encapsulate all this merge/update and load logic is cleaner.

cpascual:
Just another wild idea... maybe if we do things generic enought, we could leave the door open to also integrate the loading of .ini settings in this mechanism (not necessarily in this PR!)

lib/taurus/qt/qtgui/taurusgui/config_loader/abstract.py
stanislaw47: I've added it since some value was required to set for `TaurusGui._confDirectory`. However, I think it can be removed from abstract interface and just fallback to `$HOME` or `$HOME/.config/<ORGANIZATION>`

setup.py
stanislaw47: I was not aware of `taurus.core.util.plugin.selectEntryPoint` existence. Will change.

lib/taurus/qt/qtgui/taurusgui/config_loader/abstract.py
stanislaw47: Will be moved to `bck-compat` config loader. What about Sardana options? New loader like `sardana.something`?

stanislaw47:
I've decided to start small - now configuration is just passed around in dictionary. Each loader just creates dictionary with fields and values from configuration. Then, all created dictionaries are merged into one in order of loaded entrypoints. I believe now it is much more flexible for change and plugging in new loaders. 

I've introduced field `CONFIG_VALUES` (dunno if name is good enough) which is a list (to allow modifications, instead of tuple) of fields which can occur in configuration and be loaded. All those field are non-deprecated, Sardana-free, simple values. All `*Description` like fields are loaded manually because I do not believe that they will require such flexibility anytime soon. Custom loaders for Sardana or for backward compatibility can extend it before loading stuff.

Another thing about new config values introduced by new loaders - they have to be loaded somehow. I think that there might be space for introducing hooks in `TaurusGui.loadConfigurtation` method. Config loader could somehow register (list of) functions which would be executed during `loadConfiguration` (probably at the end). Each call would pass `TaurusGui` instance as first argument and `conf` dictionary as a second argument. That would allow flexible mutation of `TaurusGui`. I think it might be worth doing in this very PR.

lib/taurus/qt/qtgui/taurusgui/config_loader/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+
+import pkg_resources
+
+from taurus import warning
+
+__all__ = ["getLoader"]
+
+
+def getLoader(confname):
```
cpascual: Various comments regarding getLoader:

- Maybe I miss something, but I do not see any return from getLoader
- I personally prefer to only put import-related stuff in the `__init__.py` . IMHO, getLoader deserves its own file
- since it is multi-loader, I would call it `getLoaders()`

lib/taurus/qt/qtgui/taurusgui/config_loader/__main__.py
```diff
@@ -0,0 +1,56 @@
...
+        print_component(p)
+    print("AppletDescriptions: ")
+    for p in l.applets:
+        print_component(p)
+    print("ExternalApp: ")
+    for p in l.external_apps:
+        print("\t" + str(p))
+
+
+main(sys.argv[1])
```
cpascual, line 18: This makes more sense as a taurus gui option/subcommand. Also, I think that the helper functions are better provided as method(s) of the configloader object (this way different loaders can take care of printing their own stuff)

lib/taurus/qt/qtgui/taurusgui/config_loader/abstract.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Base exception raised by ConfigLoader
+    """
+
+    def __init__(self, message):
+        message = "Exception raised while loading configuration: " + message
+        super(ConfigLoaderError, self).__init__(message)
+
+
+class AbstractConfigLoader(abc.ABC):
```
cpascual, line 42: I like this implementation even more than the one before!

lib/taurus/qt/qtgui/taurusgui/config_loader/sardana.py
```diff
@@ -0,0 +1,206 @@
...
+import taurus
+from taurus.external.qt import Qt
+from taurus.qt.qtgui.taurusgui.config_loader.abstract import \
+    AbstractConfigLoader
+from taurus.qt.qtgui.taurusgui.utils import PanelDescription
+
+__all__ = ["SardanaConfigLoader"]
+
+
+class SardanaConfigLoader(AbstractConfigLoader):
```
cpascual: Ideally, this config loader should be moved to the sardana project, and the sardana module would register the plugin.
Unfortunately, this makes things a bit more complex regarding backwards-compatibility (specially because if we want to support py2 we would need to backport it to the sardana 2.x branch, or create a separate taurus-legacy-plugin project).
Apart, it also would make it a bit more difficult to test and spot regressions while developing this PR.
So I propose to keep it here until the PR is ready and tested and then consider if we can move it out of taurus before merging

lib/taurus/qt/qtgui/taurusgui/config_loader/sardana.py
```diff
@@ -0,0 +1,206 @@
...
+        "MACROSERVER_NAME",
+        "MACRO_PANELS",
+        "DOOR_NAME",
+        "MACROEDITORS_PATH",
+        "INSTRUMENTS_FROM_POOL",
+    ]
+
+    def __init__(self, confname):
+        super(SardanaConfigLoader, self).__init__(confname)
+        AbstractConfigLoader.CONFIG_VALUES + self.SARDANA_VALUES
```
cpascual: I don't know how safe is to monkeypatch the **class member**  AbstractConfigLoader.CONFIG_VALUES.

I have the impression that this may cause problems with other loaders which may not be ready to handle the injected keys.

Why not create set of less-hacky Sardana loaders? (SardanaPyConfigLoader, SardanaXmlConfigLoader, etc?) The hooks and most of the logic can be implemented in a base class that they all inherit.

lib/taurus/qt/qtgui/taurusgui/config_loader/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    Discover proper config loader based on passed string.
+    It can be either path to file or directory or Python
+    abolute path to module with configuration.
+
+    :param confname: name of configuration
+    :return: A AbstractConfigLoader subclass object
+    """
+
+    loaders = []
+    for ep in pkg_resources.iter_entry_points("taurus.gui.loaders"):
```
cpascual: Tip: to facilitate testing with entry points, I found that a good trick is to create a variable with the entry-point group name (e.g. doing):

```python
EP_GROUP_LOADERS = "taurus.gui.loaders"
for ep in pkg_resources.iter_entry_points(EP_GROUP_LOADERS):
    ...
```

In this way the tests can use the `monkeypatch`pytest  fixture to change EP_GROUP_LOADERS to some temporary value in order to have a more controlled and isolated test environment

cpascual:
note that #1114 is already merged, so if you merge develop into this branch you will already have the fixed testsuite

lib/taurus/qt/qtgui/taurusgui/config_loader/sardana.py
stanislaw47: I think that, too.

lib/taurus/qt/qtgui/taurusgui/config_loader/__init__.py
stanislaw47: Good point, will do.

lib/taurus/qt/qtgui/taurusgui/config_loader/sardana.py
stanislaw47: I admit it is very hacky but that was the simplest solution I could think of. I admit, it may not be super safe or even safe at all. From my point of view, it could work. But okay, I'll try to unhackify it.

lib/taurus/qt/qtgui/taurusgui/config_loader/__main__.py
stanislaw47: Do not know yet how exactly that would work. But I simplified this a lot since now there's internal dictionary instead of  class.

To make it simply, I needed to introduce `to_dict` method for each `*Description` class. In future, this could evolve to what was referenced as `ConfigWriter`.

lib/taurus/qt/qtgui/taurusgui/config_loader/__init__.py
stanislaw47: Done

lib/taurus/qt/qtgui/taurusgui/config_loader/abstract.py
stanislaw47: Great to hear it :)

lib/taurus/qt/qtgui/taurusgui/config_loader/sardana.py
stanislaw47: I've reimplemented it in (I hope so) safe way for py and xml. I could do this for json also but then I got to thinking: if someone will create new loader plugin (YAML/TOML), loading Sardana stuff will not work out of box. One would need to introduce extra loader (very simple on, but still) just in order to load also Sardana stuff.

Now having `CONFIG_VALUES` as class variable adds not benefit at all. It also would be very difficult to provide pluggable loading for other things like `SYNOPTIC` (which I think would be beneficial) since it also would also not work out of box for non-builtin loaders.

Previous would work for all loaders out of the box. I wonder if we can have it both ways.

From what I've seen in code, all Sardana-related values are very basic types (string, boolean) apart from `INSTRUMENTS_FROM_POOL` which will be list of strings (I think) - this may be more difficult to handle but I think *only a bit*. All current loaders can cope with those types because other fields can be of those types. Future custom loaders should also be able to load those types - this is reasonable assumption.

I would very like to have that kind of dynamic monkey patching. Class variable may not be best candidate for - agreed but it might be only way to do it *flexibly enough*.

stanislaw47: After thinking about it for a while, one of solutions I can think of is using module variable as singleton. It would keep a list of fields to extract from config source. Only problem is that config loader has to be generic enough to support extracting data of different formats simply by referencing its name. JSON and Python ones are able to do it. I believe that some others like YAML or TOML would support it also. I do not know that would be exactly implemented when it comes to other types like databases.

Problem occurs with XML - I am by no means XML expert and now it looks like extracting some complex types like lists of dictionaries is really PITA. Ha, extracting even simple ones but with *type other than string* seams to be challenging. Only think I can think of is to count on some XML-experienced Python developer help to get it right and generic enough. And implement injecting extra config values into `CONFIG_VALUES` list, be it implemented as base class variable, module's variable or anything else. 

I think now two kinds of APIs are exposed by `ConfigLoader` - one from extracting data from source and one for mutating TaurusGUI (hooks). At first sight it looks bad (`do one thing and do it right`) but after a while, I think this is the right place to do it. It would be nice to have mutating API capability to not only add but also delete or even completely override list of loaded values from configuration. That way, whole universe of different configuration options is possible. The problem with that approach is it can be not that simple - there might be some mechanism that requires changes in many places to be implemented (vide synoptic and instrument panels - all over the code). Honestly, I do know how to approach it and any help in case of architectural decisions is appreciated.

stanislaw47:
Hi @cpascual 
sorry for the long break. 
I've updated my branch with new approach to special config values (`CONSOLE` i.e.). In general, there should be generic representation of config source( be it Python module, xml or json file) and yet another entrypoint (`taurus.config.properties`). Functions registered using this entrypoint receive `TaurusGUI` instance. This allows very flexible behavior nad loading extra stuff i.e. Sardana related.
I've put `SYNOPTIC` into separate module because for me it seems like extra thing. I can revert it if this is necessary.

I would appreciate review to validate that this approach is good enough.

cpascual:
ok... give me a bit of time (I've my hands full adapting the testsuite so that it can be run with gitlab-ci). But do not hesitate to ping if I forget

PR #1117, opened 2020-06-22
LineEdit: enforce explicit disable

cpascual:
TaurusValueLineEdit auto enables/disables itself based based on events. This precludes the user
from explicitly disable it. Fix by avoiding reenabling if setEnabled(False) is called. Note that auto-disabling on errors is still allowed even if setEnabled(True) is called, since the widget
is not usable in case of error.

To test this you can use the following snippet (without this PR, the second widget will start disabled but will be autoenabled on the first change event, while with this PR it will stay disabled even if changes are forced with the first widget):

```python


if __name__ == "__main__":
    import sys

    w = Qt.QWidget()
    ly = Qt.QVBoxLayout()
    w.setLayout(ly)
    model = "sys/tg_test/1/short_scalar"

    e1 = TaurusValueLineEdit()
    e1.setModel(model)
    ly.addWidget(e1)

    e2 = TaurusValueLineEdit()
    e2.setModel(model)
    ly.addWidget(e2)
    e2.setEnabled(False)

    w.show()

    sys.exit(app.exec_())
```

### DIFF
--- a/lib/taurus/qt/qtgui/input/tauruslineedit.py
+++ b/lib/taurus/qt/qtgui/input/tauruslineedit.py
@@ -77,6 +77,7 @@ class TaurusValueLineEdit(Qt.QLineEdit, TaurusBaseWritableWidget):
         self._enableWheelEvent = False
         self._last_value = None
         self._singleStep = 1.
+        self._allow_auto_enable = True
 
         self.setAlignment(Qt.Qt.AlignRight)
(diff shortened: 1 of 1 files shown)


cpascual:
@taurus-org/integrators can you please integrate this?

teresanunez:
i have tested and it works as expected. I merge it.

PR #1146, opened 2020-09-02
Ensure that (internal) files are closed in {save,load}ConfigFile

cpascual:
The {save,load}ConfigFile methods accept either a file name or a file object
as their argument. If a name is passed, a file object is internally created
and opened, but it is not closed. This may cause trouble. Make sure the
file is closed if it was internally created.
Note that, in order to maintain backwards compatibility, the file
won't be closed if it was passed already as a file object (the
responsibility for closing lies in the opener)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,9 @@ matrix:
     - env: TOXENV=py35-qt5
     - env: TOXENV=py36-qt5
     - env: TOXENV=py37-qt5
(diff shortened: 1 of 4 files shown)


cpascual:
@taurus-org/integrators : this is ready for review

Note, the last 2 commits are cherry picks from develop added to fix the CI tests (due to unrelated issues). Only 813f255 needs to be reviewed.

PR #1151, opened 2020-09-14
Avoid warning on strings with replacement fields

### LINKED ISSUE #1149: TaurusLabel: Warning when {} is part of the attribute value
Hi,

Here is a log i receive from a device.
```
MainThread     WARNING  2020-09-09 20:13:35 b'TaurusLabel.foo/metadata/manager/dataFolder': Error formatting display (KeyError('scanName',)). Reverting to raw string
```
The short story is that the device attribute can contain strings like that `/foo/bar/{scanName}`.

Do you consider this warning relevant?

This is due to this part of the code. But `value` here is already pre-formatted. So there is not anymore distinction between the data value and the formatting. I feel like it's not easy to fix.
```
        try:
            v = value.format(dev=dev, attr=attr)
        except Exception as e:
            self.warning(
                "Error formatting display (%r). Reverting to raw string", e)
            v = value
```

I think i will only skip it on my logging config.
But just to let you know it can happen.

cpascual:
TaurusLabel.displayValue fails when displaying a string attribute
whose value contains a replacement field, and falls back to the
default representation with a warning.
Avoid this by using a custom string formatter that ignores
missing keys when formatting a string.
Thanks to @vallsv for the proposed solution.

Fixes #1149

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,9 @@ matrix:
     - env: TOXENV=py35-qt5
     - env: TOXENV=py36-qt5
     - env: TOXENV=py37-qt5
+    - env: TOXENV=py38-qt5
     - env: TOXENV=py37-ps2
-    - env: TOXENV=py37-qt5-docs
+    - env: TOXENV=py38-qt5-docs
   allow_failures:
     #- env: TOXENV=flake8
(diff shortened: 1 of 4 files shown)


vallsv:
I can confirm it works.

PR #1154, opened 2020-09-30
Fix buffer option in taurus trend cli

cpascual:
An indentation mistake provokes a premature exit when using the --buffer option in the taurus trend CLI (even on success)
Fix it so that it only exits in case of error.

Also move the call to setting the buffers in the CLI command after the models have been set, in order to affect them

### DIFF
--- a/lib/taurus/cli/alt.py
+++ b/lib/taurus/cli/alt.py
@@ -328,7 +328,7 @@ def trend_cmd(
                 epname,
                 e,
             )
-        sys.exit(1)
+            sys.exit(1)
 
     # set models
     if models:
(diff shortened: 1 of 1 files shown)


cpascual:
@taurus-org/integrators : This is ready for review. In order to test it with taurus_pyqtgraph trends,  https://github.com/taurus-org/taurus_pyqtgraph/pull/47 is required

cpascual:
taurus-org/taurus_pyqtgraph#47 is already merged. 
can any @taurus-org/integrators have a look at this?

teresanunez (approved):

cpascual:
Thnks @teresanunez for the review. You marked this as approved. Does this mean that it can be merged? if so, I can do it, but you can do it directly too.

